PR #1, opened 2019-01-21
Add Resource Manangement Service

jjbuchan:
# Resolves

https://jira.rax.io/browse/SALUS-121 - Create Resource Management Service
https://jira.rax.io/browse/SALUS-137 - Resource Management HTTP Interface
https://jira.rax.io/browse/SALUS-138 - Namespace envoy labels

# What

The initial implementation of the Resource Management service.

Adds a REST API service as well as the intermediate service to perform SQL operations.

The API performs CRUD operations for Resource objects.  It has also been configured with custom pagination/error handling to allow for easier integration with Intelligence.

For these requests, the resource management service posts new messages to the resources topic in Kafka for consumption by the Monitoring Management and Presence Monitoring services.  Additionally it consumes new attach events from Kafka to register those resources in SQL.

# How

* `hibernate-jpamodelgen` was added to `salus-telemetry-model` to allow us to query the resources via JPA Metamodel data rather than writing raw sql.
  * This ties in to how we use the CriteriaBuilder in `ResourceManagement.java`
* New `ResourceCreate` and `ResourceUpdate` models were introduced to represent the valid payloads to be received by the API.
* Custom error handling was added in `RestExceptionHandler.java`
* `ApplicationEventPublisher` is used in the API to send events to a listener to add additional header data to the responses to assist with pagination handling by the clients.  These listeners perform their actions synchronously.
* Tests are ran in an embedded h2 database via the inclusion of the `com.h2database` dependency.
* Test data can be auto-populated via `uk.co.jemos.podam`
    * We currently do have to use a deprecated version of `@NotBlank` for this to work.
    * I will look into adding support for the newer version to that project.

## How to test

Currently `ResourceManagementTest.java` can be run.  More API and Kafka tests will be coming in the next PR.

# Why

This service will be queried by the public API as well as by other services to get Resource information.  It is the source of truth for Resource data.

It has to correlate new envoy attachments to existing resources created via the public API and vice-versa.

# TODO
- Add tests
- General cleanup
- Possibly have to add a default sort value to all queries.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1,0 +1,108 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>org.springframework.boot</groupId>
+    <artifactId>spring-boot-starter-parent</artifactId>
+    <version>2.1.1.RELEASE</version>
+    <relativePath/> <!-- lookup parent from repository -->
+  </parent>
+  <groupId>com.rackspace.salus</groupId>
+  <artifactId>salus-telemetry-resource-management</artifactId>
+  <version>0.0.1-SNAPSHOT</version>
+  <name>salus-telemetry-resource-management</name>
+  <description>Service for handling Resource CRUD events</description>
+
+  <properties>
+    <java.version>1.8</java.version>
+  </properties>
+
+  <dependencies>
+    <dependency>
+      <groupId>org.springframework.boot</groupId>
+      <artifactId>spring-boot-starter</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.springframework.kafka</groupId>
+      <artifactId>spring-kafka</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.springframework.boot</groupId>
+      <artifactId>spring-boot-starter-json</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.springframework.boot</groupId>
+      <artifactId>spring-boot-devtools</artifactId>
+      <scope>runtime</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.springframework.boot</groupId>
+      <artifactId>spring-boot-starter-data-jpa</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>mysql</groupId>
+      <artifactId>mysql-connector-java</artifactId>
+      <scope>runtime</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.projectlombok</groupId>
+      <artifactId>lombok</artifactId>
+      <optional>true</optional>
+    </dependency>
+    <dependency>
+      <groupId>org.springframework.boot</groupId>
+      <artifactId>spring-boot-starter-test</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.springframework.kafka</groupId>
+      <artifactId>spring-kafka-test</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>com.fasterxml.jackson.core</groupId>
+      <artifactId>jackson-databind</artifactId>
+      <version>2.9.7</version>
+      <scope>compile</scope>
+    </dependency>
+    <dependency>
+      <groupId>com.rackspace.salus</groupId>
+      <artifactId>salus-telemetry-protocol</artifactId>
+      <version>0.1.0-SNAPSHOT</version>
+      <scope>compile</scope>
+    </dependency>
+    <dependency>
+      <groupId>com.rackspace.salus</groupId>
+      <artifactId>salus-telemetry-model</artifactId>
+      <version>0.1.0-SNAPSHOT</version>
+      <scope>compile</scope>
+    </dependency>
+    <dependency>
+      <groupId>com.rackspace.salus</groupId>
+      <artifactId>salus-telemetry-protocol</artifactId>
+      <version>0.1.0-SNAPSHOT</version>
+      <scope>compile</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.hibernate</groupId>
+      <artifactId>hibernate-core</artifactId>
+      <version>5.3.7.Final</version>
+    </dependency>
+    <dependency>
+      <groupId>commons-beanutils</groupId>
+      <artifactId>commons-beanutils</artifactId>
+      <version>1.9.3</version>
+    </dependency>
+  </dependencies>
+
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.springframework.boot</groupId>
+        <artifactId>spring-boot-maven-plugin</artifactId>
+      </plugin>
+    </plugins>
+  </build>
+
+</project>
--- a/pom.xml
+++ b/pom.xml
@@ -76,7 +76,12 @@
     <dependency>
       <groupId>uk.co.jemos.podam</groupId>
       <artifactId>podam</artifactId>
-      <version>LATEST</version>
+      <version>7.2.1.RELEASE</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.httpcomponents</groupId>
+      <artifactId>httpclient</artifactId>
       <scope>test</scope>
     </dependency>
   </dependencies>
--- a/pom.xml
+++ b/pom.xml
@@ -20,21 +20,12 @@
 
   <dependencies>
     <dependency>
-      <groupId>org.springframework.boot</groupId>
-      <artifactId>spring-boot-starter</artifactId>
-    </dependency>
-    <dependency>
       <groupId>org.springframework.kafka</groupId>
       <artifactId>spring-kafka</artifactId>
     </dependency>
     <dependency>
       <groupId>org.springframework.boot</groupId>
-      <artifactId>spring-boot-starter-json</artifactId>
-    </dependency>
-    <dependency>
-      <groupId>org.springframework.boot</groupId>
-      <artifactId>spring-boot-devtools</artifactId>
-      <scope>runtime</scope>
+      <artifactId>spring-boot-starter-web</artifactId>
     </dependency>
     <dependency>
       <groupId>org.springframework.boot</groupId>
@@ -61,12 +52,6 @@
       <scope>test</scope>
     </dependency>
     <dependency>
-      <groupId>com.fasterxml.jackson.core</groupId>
-      <artifactId>jackson-databind</artifactId>
-      <version>2.9.7</version>
-      <scope>compile</scope>
-    </dependency>
-    <dependency>
       <groupId>com.rackspace.salus</groupId>
       <artifactId>salus-telemetry-protocol</artifactId>
       <version>0.1.0-SNAPSHOT</version>
@@ -79,20 +64,20 @@
       <scope>compile</scope>
     </dependency>
     <dependency>
-      <groupId>com.rackspace.salus</groupId>
-      <artifactId>salus-telemetry-protocol</artifactId>
-      <version>0.1.0-SNAPSHOT</version>
-      <scope>compile</scope>
-    </dependency>
-    <dependency>
-      <groupId>org.hibernate</groupId>
-      <artifactId>hibernate-core</artifactId>
-      <version>5.3.7.Final</version>
-    </dependency>
-    <dependency>
       <groupId>commons-beanutils</groupId>
       <artifactId>commons-beanutils</artifactId>
       <version>1.9.3</version>
+    </dependency>
+    <dependency>
+      <groupId>com.h2database</groupId>
+      <artifactId>h2</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>uk.co.jemos.podam</groupId>
+      <artifactId>podam</artifactId>
+      <version>LATEST</version>
+      <scope>test</scope>
     </dependency>
   </dependencies>
 
--- a/src/main/java/com/rackspace/salus/resource_management/services/KafkaEgress.java
+++ b/src/main/java/com/rackspace/salus/resource_management/services/KafkaEgress.java
@@ -14,10 +14,12 @@
  * limitations under the License.
  */
 
-package com.rackspace.salus.resource_management.services.services;
+package com.rackspace.salus.resource_management.services;
 
-import com.rackspace.salus.resource_management.services.config.ResourceManagementProperties;
-import com.rackspace.salus.resource_management.services.types.KafkaMessageType;
+import com.rackspace.salus.resource_management.config.ResourceManagementProperties;
+import com.rackspace.salus.telemetry.messaging.KafkaMessageType;
+import com.rackspace.salus.telemetry.messaging.ResourceEvent;
+import com.rackspace.salus.telemetry.model.Resource;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.kafka.core.KafkaTemplate;
 import org.springframework.stereotype.Service;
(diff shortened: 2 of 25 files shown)


pom.xml
```diff
@@ -0,0 +1,93 @@
...
+    </dependency>
+    <dependency>
+      <groupId>com.h2database</groupId>
+      <artifactId>h2</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>uk.co.jemos.podam</groupId>
+      <artifactId>podam</artifactId>
+      <version>LATEST</version>
```
itzg: Please call out the specific version (7.2.1.RELEASE) since I'm surprised Maven 3.x is still allowing `LATEST`.

src/main/java/com/rackspace/salus/resource_management/services/KafkaEgress.java
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    public void sendResourceEvent(ResourceEvent event) {
+        final String topic = properties.getKafkaTopics().get(KafkaMessageType.RESOURCE);
+        if (topic == null) {
+            throw new IllegalArgumentException(String.format("No topic configured for %s", KafkaMessageType.RESOURCE));
+        }
+
+        Resource resource = event.getResource();
+        String key = String.format("%s:%s", resource.getTenantId(), resource.getResourceId());
```
itzg: Maybe `KafkaMessageKeyBuilder` should also have a helper method for composing keys with arbitrary values just to keep the key composition logic all in one place.

I went ahead and created https://github.com/racker/salus-common/pull/8 for this.

src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
```diff
@@ -0,0 +1,363 @@
...
+    public List<Resource> getResources(String tenantId, Map<String, String> labels) {
+        // use geoff's label search query
+    }*/
+
+    /**
+     * Get all resources where the presence monitoring field matches the parameter provided.
+     * @param presenceMonitoringEnabled Whether presence monitoring is enabled or not.
+     * @return Stream of resources.
+     */
+    public Stream<Resource> getResources(boolean presenceMonitoringEnabled) {
```
itzg: Remind me, why would a resource not have presence monitoring enabled? A poller resource?

src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
```diff
@@ -0,0 +1,363 @@
...
+     * Registers or updates resources in the datastore.
+     * Prefixes the labels received from the envoy so they do not clash with any api specified values.
+     *
+     * @param attachEvent The event triggered from the Ambassador by any envoy attachment.
+     */
+    public void handleEnvoyAttach(AttachEvent attachEvent) {
+        String tenantId = attachEvent.getTenantId();
+        String resourceId = attachEvent.getResourceId();
+        Map<String, String> labels = attachEvent.getLabels();
+        labels = applyNamespaceToKeys(labels, "envoy");
```
itzg: We could eventually have a `LabelNamespace` class or enum somewhere (like model or common) that declares the label namespaces. For now, please just declare a constant in this class for "envoy".

src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
```diff
@@ -0,0 +1,363 @@
...
+     * @param resource The resource to update.
+     * @param envoyLabels The list of labels received from a newly connected envoy.
+     */
+    private void updateEnvoyLabels(Resource resource, Map<String, String> envoyLabels) {
+        AtomicBoolean updated = new AtomicBoolean(false);
+        Map<String, String> resourceLabels = resource.getLabels();
+        Map<String, String> oldLabels = new HashMap<>(resourceLabels);
+
+        oldLabels.forEach((name, value) -> {
+            if (name.startsWith("envoy.")) {
```
itzg: If you want to do this with some more Java 8 coolness, you could do a filter via a stream of the map entries:

```
    oldLabels.entrySet().stream()
        .filter(entry -> entry.getKey().startsWith("envoy."))
        .forEach(entry -> {
           // same as your existing block
        });
```

src/main/java/com/rackspace/salus/resource_management/web/controller/ResourceApi.java
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    @GetMapping("/resources")
+    public List<Resource> getAll(@RequestParam(defaultValue = "100") int size,
+                                 @RequestParam(defaultValue = "0") int page,
+                                 UriComponentsBuilder uriBuilder,
+                                 final HttpServletResponse response) {
+
+        Page<Resource> result = resourceManagement.getAllResources(PageRequest.of(page, size));
+        eventPublisher.publishEvent(new PaginatedResultsRetrievedEvent<>(
```
itzg: Depending on the event handlers to mutate the given fields, such as the `HttpServletResponse`, feels wrong. I would prefer to see the usage of `PaginatedResultsRetrievedDiscoverabilityListener` be explicitly invoked via an autowiring of that.

src/main/java/com/rackspace/salus/resource_management/web/listener/PaginatedResultsRetrievedDiscoverabilityListener.java
```diff
@@ -0,0 +1,125 @@
...
+import org.springframework.context.ApplicationListener;
+import org.springframework.stereotype.Component;
+import org.springframework.web.util.UriComponentsBuilder;
+
+import com.google.common.base.Preconditions;
+import com.google.common.net.HttpHeaders;
+
+@SuppressWarnings({ "rawtypes" })
+@Component
+class PaginatedResultsRetrievedDiscoverabilityListener implements ApplicationListener<PaginatedResultsRetrievedEvent> {
```
itzg: I haven't explored HATEOAS much at all, but you might have found https://spring.io/projects/spring-hateoas to be a helpful solution. I'm usually too lazy to do these proper resource references in the REST API responses, so take my suggestion with a grain of salt.

I'm curious, did SSE require these response details to be populated?

src/main/resources/application.yml
```diff
@@ -0,0 +1,34 @@
...
+    LOG:
+      telemetry.logs.json
+    METRIC:
+      telemetry.metrics.json
+    EVENT:
+      telemetry.events.json
+    ATTACH:
+      attaches
+    RESOURCE:
+      resources
```
itzg: In UMB we have started to use the topic naming convention of "{namespace}.{thing_in_namespace}.{payload_type}". Please devise topic names that fit the convention, where `{namespace}` will be "telemetry" since that's short for Salus Telemetry.

src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
jjbuchan: Firewalls, loadbalancers... things that don't have an envoy.

Also, based on our experience with ele, it could be good to be able to turn if off for anything.

src/main/java/com/rackspace/salus/resource_management/web/controller/ResourceApi.java
jjbuchan: I did it this way after reading some of the articles related to this repo - https://github.com/eugenp/tutorials/tree/master/spring-rest-full from our friends at baeldung.

src/main/java/com/rackspace/salus/resource_management/web/listener/PaginatedResultsRetrievedDiscoverabilityListener.java
jjbuchan: Sse doesn't use all this stuff, but the more public facing endpoints do.  My assumption is that the graphql api will need to be able to expose these details.  I know the lack of those details is one of the problems intelligence currently has with ele.

src/main/resources/application.yml
jjbuchan: I was talking to george yesterday and it seems odd that we need to define these topics in each repo.  Should this be something that is somehow shared?

src/main/java/com/rackspace/salus/resource_management/web/controller/ResourceApi.java
itzg: Indeed they did and the approach makes more sense glancing at

https://www.baeldung.com/rest-api-discoverability-with-spring

I'm surprised they don't touch on the Spring HATEOAS project there though; however, maybe that Spring project is not prime time yet.

src/main/resources/application.yml
itzg: I know, I have been wondering the same. One thought was that maybe we just need a general `KafkaTopicsProperties` bean with a `@ConfigurationPropertes("kafka-topics")` that we can pull into our apps. Some of the redundancy though will be negated by some shared kubernetes config (like a common ConfigMap) in production, etc.

jjbuchan: Updated it here for now.

src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
jjbuchan: Yeah, that type of thing was on my mind.  Went for the lazy approach to start with.

src/main/java/com/rackspace/salus/resource_management/services/KafkaEgress.java
jjbuchan: I'll switch in the next PR.

pom.xml
jjbuchan: Changed

src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
```diff
@@ -0,0 +1,363 @@
...
+                    resourceId, tenantId));
+        }
+        Map<String, String> oldLabels = new HashMap<>(resource.getLabels());
+        boolean presenceMonitoringStateChange = false;
+        if (updatedValues.getPresenceMonitoringEnabled() != null) {
+            presenceMonitoringStateChange = resource.getPresenceMonitoringEnabled().booleanValue()
+                    != updatedValues.getPresenceMonitoringEnabled().booleanValue();
+        }
+
+        PropertyMapper map = PropertyMapper.get();
```
jjbuchan: @itzg fyi, the previous method I found never ended up working so I went back to this.  Works well when we only have a couple fields to deal with.

itzg: Yeah, I noticed you used the Spring utility there and agree with a few fields it looks good.

src/main/java/com/rackspace/salus/resource_management/web/controller/ResourceApi.java
```diff
@@ -0,0 +1,149 @@
...
+                Resource.class, uriBuilder, response, page, result.getTotalPages(), size));
+        return result.getContent();
+    }
+
+    @GetMapping("/envoys")
+    public SseEmitter getAllWithPresenceMonitoringAsStream() {
+        SseEmitter emitter = new SseEmitter();
+        ExecutorService nonBlockingService = Executors.newCachedThreadPool();
```
GeorgeJahad: It seems a bit odd that a new thread pool is created each time this api is called.

src/main/java/com/rackspace/salus/resource_management/web/controller/ResourceApi.java
```diff
@@ -0,0 +1,149 @@
...
+    public Resource getByResourceId(@PathVariable String tenantId,
+                                    @PathVariable String resourceId,
+                                    final HttpServletResponse response) throws NotFoundException {
+
+        Resource resource = resourceManagement.getResource(tenantId, resourceId);
+        if (resource == null) {
+            throw new NotFoundException(String.format("No resource found for %s on tenant %s",
+                    resourceId, tenantId));
+        }
+        eventPublisher.publishEvent(new SingleResourceRetrievedEvent(this, response));
```
GeorgeJahad: Out of curiosity, why publish an event when a resource is retrieved?

GeorgeJahad: hmm, i assumed an event was a kafka message.  i guess that is not the case.

src/main/java/com/rackspace/salus/resource_management/web/controller/ResourceApi.java
jjbuchan: For how infrequently it'll likely be called this seemed like the best option.  There are drawbacks to using it like this or by declaring it in the constructor.  Had this been a public api method, I'd have maybe chosen the other way.

src/main/java/com/rackspace/salus/resource_management/web/controller/ResourceApi.java
jjbuchan: Nope, not a kafka event.  This is what I mentioned in the description:

> ApplicationEventPublisher is used in the API to send events to a listener to add additional header data to the responses to assist with pagination handling by the clients. These listeners perform their actions synchronously.

@itzg commented on some of this earlier as well, but after reading the tutorial I followed he agreed with the decision (I think).

src/main/java/com/rackspace/salus/resource_management/web/controller/ResourceApi.java
itzg: Threads and thread pools are fairly expensive, so it would be good to do more like [the article](https://www.baeldung.com/spring-mvc-sse-streams) and create a pool once for the `ResourceApi` instance to use across the API calls.

jjbuchan: That's how I originally did it, but that doesn't mention anything about shutting it down, and I had read it can be bad to leave it running.

Is this that expensive if it is only being used when a presence monitor service starts up?

src/main/java/com/rackspace/salus/resource_management/web/controller/ResourceApi.java
itzg: Yes, you correctly followed the approach they used with the events, but IMHO it was an odd way for them to abstract the behavior :).

src/main/java/com/rackspace/salus/resource_management/web/controller/ResourceApi.java
itzg: This is another case (the application events being the other) where Baeldung is letting me down. The proper way to create a Spring managed task executor (aka thread pool) is like this:

https://github.com/racker/salus-telemetry-ambassador/blob/1cba97463667fd6c099ae5360975b2f12c39fd3c/src/main/java/com/rackspace/salus/telemetry/ambassador/config/AsyncConfig.java#L29-L32

Being a Spring bean, Spring will take care of closing it out when the application context closes.

With that, you can just autowire the `TaskExecutor` into the API class.

itzg:
It finally dawned on me that I usually paired up `Pageable` on the request side with [`Page`](https://docs.spring.io/spring-data/commons/docs/current/api/org/springframework/data/domain/Page.html) for the responses. I also didn't realize that both of those are provided by Spring Data and not plain old Spring MVC. Would `Page` be sufficient from what you can tell?

src/main/java/com/rackspace/salus/resource_management/web/controller/ResourceApi.java
```diff
@@ -0,0 +1,148 @@
...
+
+    @Autowired
+    public ResourceApi(ResourceManagement resourceManagement, ApplicationEventPublisher eventPublisher, TaskExecutor taskExecutor) {
+        this.resourceManagement = resourceManagement;
+        this.eventPublisher = eventPublisher;
+        this.taskExecutor = taskExecutor;
+    }
+
+    @GetMapping("/resources")
+    public List<Resource> getAll(@RequestParam(defaultValue = "100") int size,
```
itzg: While researching my other comment I was able to confirm that you can actually request `Pageable` as a controller method argument:

https://docs.spring.io/spring-data/data-commons/docs/current/reference/html/#core.web

In that section I see it even mentions integrations with Spring HATEOAS.

src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
jjbuchan: Made it a constant for now.

src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
jjbuchan: Done

PR #3, opened 2019-01-23
Add API Tests

jjbuchan:
# What

Adds the tests for the resource management REST API.

It also tweaks the error handling so that json objects are returned instead of plain text.

# How

Uses a `MockBean` for the resource management service.  Since we already have tests that verify the behavior of it, we mock the responses that the API layer will receive from it, and instead just verify the API inputs/outputs are accurate based on the mock data.

## How to test

Run `ResourceApiTest`

# Why

Tests are good.

# TODO
- Add more tests
- Possibly have to add a default sort value to all queries.


pom.xml
```diff
@@ -84,6 +84,18 @@
       <artifactId>httpclient</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
```
itzg: Doesn't hurt to add them here, but both hamcrest and json-path are already pulled in by spring-boot-starter-test

https://mvnrepository.com/artifact/org.springframework.boot/spring-boot-starter-test/2.1.1.RELEASE

src/main/java/com/rackspace/salus/resource_management/web/error/RestExceptionHandler.java
```diff
@@ -34,16 +34,16 @@
     @ExceptionHandler(value = { IllegalArgumentException.class, IllegalStateException.class, ResourceAlreadyExists.class})
     protected ResponseEntity<Object> handleBadRequest(
             RuntimeException ex, WebRequest request) {
-        String errorMessage = ex.getMessage();
-        return handleExceptionInternal(ex, errorMessage,
+        GenericError error = new GenericError(ex.getMessage());
+        return handleExceptionInternal(ex, error,
```
itzg: Oh, it wasn't happy with just a string? Is it because it's a response **entity** exception handler?

src/test/java/com/rackspace/salus/resource_management/ResourceApiTest.java
```diff
@@ -0,0 +1,210 @@
...
+
+    private ObjectMapper objectMapper = new ObjectMapper();
+
+    @Test
+    public void testGetResource() throws Exception {
+        Resource resource = podamFactory.manufacturePojo(Resource.class);
+        when(resourceManagement.getResource(anyString(), anyString()))
+                .thenReturn(resource);
+
+        String tenantId = RandomStringUtils.randomAlphabetic( 8 );
```
itzg: I wonder if podam exposes a utility function for just a random string?

src/test/java/com/rackspace/salus/resource_management/ResourceApiTest.java
```diff
@@ -0,0 +1,210 @@
...
+    @MockBean
+    ResourceManagement resourceManagement;
+
+    @MockBean
+    EntityManagerFactory entityManagerFactory;
+
+    @MockBean
+    ResourceRepository resourceRepository;
+
+    private ObjectMapper objectMapper = new ObjectMapper();
```
itzg: You might actually be able to `@Autowire` one of these since the MVC test slice probably registers one.

src/main/java/com/rackspace/salus/resource_management/web/error/RestExceptionHandler.java
jjbuchan: It was happy; I wasn't happy.  It would only ever return a text response, not json.  For json you have to wrap it in an object.

itzg: That makes sense now. As you probably have noticed I have had to wrap some single field values in a "response" class for that kind of reason.

src/test/java/com/rackspace/salus/resource_management/ResourceApiTest.java
jjbuchan: That did cross my mind.  By the time I got here I was too tired and just pushed it.  I'll see if I can tweak that.

src/test/java/com/rackspace/salus/resource_management/ResourceApiTest.java
jjbuchan: I don't think they do, but I should read more of their docs.

jjbuchan:
I've simplified the pagination by removing all those listeners and just returning `Page`.  Also added more tests.

Classing this PR as done but the resource management service will still need kafka tests.

src/test/java/com/rackspace/salus/resource_management/ResourceApiTest.java
```diff
@@ -0,0 +1,335 @@
...
+                numberOfResources);
+
+        when(resourceManagement.getResources(anyString(), any()))
+                .thenReturn(pageOfResources);
+
+        String tenantId = RandomStringUtils.randomAlphabetic( 8 );
+        String url = String.format("/api/tenant/%s/resources", tenantId);
+
+        mockMvc.perform(get(url).contentType(MediaType.APPLICATION_JSON))
+                .andExpect(status().isOk())
```
itzg: It doesn't look like you're asserting the actual contents of the resources returned. It might just take an `equalTo` check of the `pageOfResources` that the mock is providing to the API layer. If that comparison work, it might eliminate the need for the individual jsonPath assertions.

jjbuchan: I don't think the comparison will work, but I'll get something added.

jjbuchan: You were right.  it works.

jjbuchan: I still like asserting on the json stuff so we know that is matches what a consumer of the api would expect

itzg: Good point since good unit tests can also serve as a form of documentation

PR #5, opened 2019-02-07
Parent the pom to the salus-app-base

itzg:
# Resolves

Discovered while looking up MySQL version needed for \SALUS-93

# What

Updates the pom's parent to match our other apps.

# How

Use the salus-app-base as the parent since that standardizes the Spring Boot version used, etc.


itzg:
@danhickox this is probably needed for the image build of this app

PR #6, opened 2019-02-08
Re-configure default server port only for dev Spring profile

itzg:
# Resolves

Part of SALUS-188

# What

Fixing up a convention I had incorrectly started with setting the `server.port` in the main application.yml. Instead we should only override the default 8080 with the `dev` profile activated. Using 8080 keeps the deployment specs consistent.

### DIFF
--- a/src/main/resources/application.yml
+++ b/src/main/resources/application.yml
@@ -1,5 +1,3 @@
-server:
-  port: 8085
 resource-management:
   kafka-topics:
     LOG:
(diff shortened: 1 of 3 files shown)


src/main/resources/application.yml
```diff
@@ -1,5 +1,3 @@
-server:
```
jjbuchan: Oops, didnt mean to leave it like this.

itzg: No problem at all. I started the pattern with some of mine.

PR #9, opened 2019-02-27
Create ResourceManagement label matching query

harageth:
# Resolves

    SALUS-124 (https://jira.rax.io/browse/SALUS-124)

# What

    Create a query that will match to all of the labels we have provided

# How

    It creates a query function to manually create the SQL query since this is a fairly complex query

## How to test

   I sent API requests to create objects with multiple different labels under different accounts to make sure I limited to the current tenant and matched the query to specific labels.

# Why

    This query incurs fewer JOIN penalties from the suggested SO approach.

# TODO

    Clean up the code. Write unit tests.


src/main/java/com/rackspace/salus/resource_management/web/controller/ResourceApi.java
```diff
@@ -115,4 +121,49 @@ public void delete(@PathVariable String tenantId,
...
+    @GetMapping("/tenant/{tenantId}/resources/{labels}")
+    public Resource getResourcesWithLabels(@PathVariable String tenantId,
+                                                @PathVariable Map<String, String> labels) {
+
+        Query query = constructQuery(labels, tenantId);
+        query.getResultList();
+        return null;
+    }
+
+    private Query constructQuery(Map<String, String> labels, String tenantId) {
```
GeorgeJahad: Is there a reason why this is not in the resourceManagement class with all the other resource db code?

src/main/java/com/rackspace/salus/resource_management/web/controller/ResourceApi.java
```diff
@@ -115,4 +121,49 @@ public void delete(@PathVariable String tenantId,
...
         resourceManagement.removeResource(tenantId, resourceId);
     }
+
+    @GetMapping("/tenant/{tenantId}/resources/{labels}")
+    public Resource getResourcesWithLabels(@PathVariable String tenantId,
+                                                @PathVariable Map<String, String> labels) {
+
+        Query query = constructQuery(labels, tenantId);
+        query.getResultList();
+        return null;
```
GeorgeJahad: why is this returning null?

harageth: Oh I must not have saved that change when I added the file.

src/main/java/com/rackspace/salus/resource_management/web/controller/ResourceApi.java
harageth: No good reason. I started it there and then just didn't think about moving it. 

Moving it now

GeorgeJahad:
> Clean up the code. Write unit tests. 
For WIP prs, I prefer that they have WIP in the title so we know they aren't really ready for review yet.

src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
```diff
@@ -361,5 +364,39 @@ private void removePresenceMonitoring(String tenantId, String resourceId) {
...
+
+        String query = "SELECT * FROM resources WHERE id IN (SELECT id FROM resources id IN ";
+        StringBuilder builder = new StringBuilder(query);
+        builder.append("SELECT id from resource_labels WHERE id IN ( SELECT id FROM resources WHERE tenant_id = :tenant_id) AND ");
+
+        int i = 0;
+        labels.size();
+        for(Map.Entry<String, String> entry : labels.entrySet()) {
+            if(i > 0) {
+                builder.append(" OR ");
```
itzg: The label tuples need to be AND'ed . This is the code you're replacing:

https://github.com/racker/salus-telemetry-etcd-adapter/blob/master/src/main/java/com/rackspace/salus/telemetry/etcd/services/EnvoyLabelManagement.java#L79-L101

src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
```diff
@@ -361,5 +364,39 @@ private void removePresenceMonitoring(String tenantId, String resourceId) {
         saveAndPublishResource(resource, resource.getLabels(), true, OperationType.UPDATE);
     }
 
+    public Query constructQuery(Map<String, String> labels, String tenantId) {
```
itzg: Please name this method something that indicates the business logic being performed.

GeorgeJahad:
FYI:  I tried running this WIP, I needed to do the following to get the request parameters set correctly:
https://github.com/racker/salus-telemetry-resource-management/compare/6c64efcd5f050eeff8dc7171530f024983b03c87...george/fixApi

Otherwise, the mvc router tried to route it to one of the other handlers, and the labels var got set with the path variables, (tenantId), rather than the request parameters.

With that change, I can see the parameters and routing work with this call:
curl  'localhost:8085/api/tenant/aaaaa/resourceLabels?os=LINUX&arch=X86'

However the following error is returned.  I don't know sql well enough to understand the problem.

{"message":"org.hibernate.hql.internal.ast.QuerySyntaxException: unexpected token: * near line 1, column 8 [SELECT * FROM resources WHERE id IN (SELECT id FROM resources id IN SELECT id from resource_labels WHERE id IN ( SELECT id FROM resources WHERE tenant_id = :tenant_id) AND labels = :label0 AND labels_key = :key0 OR labels = :label1 AND labels_key = :key1) GROUP BY id HAVING COUNT(id) = :i]"}

itzg:
I found [an SO answer](https://stackoverflow.com/a/1965576/121324) that gave some clues about logging the EntityManager/JPA transactions and it looks like normally the transaction is wrapping just the repository save operation when an Envoy attachment event is processed. The reason the unit test behaves differently is because the transaction is `PROPAGATION_REQUIRED` and the unit test itself started a broader transaction.

Below is the logs I saw when setting this in my `application-dev.yml`:
```
logging:
  level:
    org.hibernate.type: trace
    org.springframework.orm.jpa: debug
```

```
2019-03-13 08:05:54.686 DEBUG 4613 --- [ntainer#0-0-C-1] o.s.orm.jpa.JpaTransactionManager        : Creating new transaction with name [org.springframework.data.jpa.repository.support.SimpleJpaRepository.save]: PROPAGATION_REQUIRED,ISOLATION_DEFAULT
2019-03-13 08:05:54.686 DEBUG 4613 --- [ntainer#0-0-C-1] o.s.orm.jpa.JpaTransactionManager        : Opened new EntityManager [SessionImpl(1440863985<open>)] for JPA transaction
2019-03-13 08:05:54.689 DEBUG 4613 --- [ntainer#0-0-C-1] o.s.orm.jpa.JpaTransactionManager        : Exposing JPA transaction as JDBC [org.springframework.orm.jpa.vendor.HibernateJpaDialect$HibernateConnectionHandle@6ba0d4b1]
Hibernate: select resource0_.id as id1_3_0_, resource0_.presence_monitoring_enabled as presence2_3_0_, resource0_.resource_id as resource3_3_0_, resource0_.tenant_id as tenant_i4_3_0_ from resources resource0_ where resource0_.id=?
2019-03-13 08:05:54.690 TRACE 4613 --- [ntainer#0-0-C-1] o.h.type.descriptor.sql.BasicBinder      : binding parameter [1] as [BIGINT] - [10]
2019-03-13 08:05:54.691 TRACE 4613 --- [ntainer#0-0-C-1] o.h.type.descriptor.sql.BasicExtractor   : extracted value ([presence2_3_0_] : [BOOLEAN]) - [true]
2019-03-13 08:05:54.691 TRACE 4613 --- [ntainer#0-0-C-1] o.h.type.descriptor.sql.BasicExtractor   : extracted value ([resource3_3_0_] : [VARCHAR]) - [development:0]
2019-03-13 08:05:54.691 TRACE 4613 --- [ntainer#0-0-C-1] o.h.type.descriptor.sql.BasicExtractor   : extracted value ([tenant_i4_3_0_] : [VARCHAR]) - [aaaaaa]
2019-03-13 08:05:54.692 TRACE 4613 --- [ntainer#0-0-C-1] org.hibernate.type.CollectionType        : Created collection wrapper: [com.rackspace.salus.telemetry.model.Resource.labels#10]
Hibernate: select labels0_.id as id1_2_0_, labels0_.labels as labels2_2_0_, labels0_.labels_key as labels_k3_0_ from resource_labels labels0_ where labels0_.id=?
2019-03-13 08:05:54.692 TRACE 4613 --- [ntainer#0-0-C-1] o.h.type.descriptor.sql.BasicBinder      : binding parameter [1] as [BIGINT] - [10]
2019-03-13 08:05:54.693 TRACE 4613 --- [ntainer#0-0-C-1] o.h.type.descriptor.sql.BasicExtractor   : extracted value ([id1_2_0_] : [BIGINT]) - [10]
2019-03-13 08:05:54.694 TRACE 4613 --- [ntainer#0-0-C-1] o.h.type.descriptor.sql.BasicExtractor   : extracted value ([labels2_2_0_] : [VARCHAR]) - [X86_64]
2019-03-13 08:05:54.694 TRACE 4613 --- [ntainer#0-0-C-1] o.h.type.descriptor.sql.BasicExtractor   : extracted value ([labels_k3_0_] : [VARCHAR]) - [envoy.arch]
2019-03-13 08:05:54.694 TRACE 4613 --- [ntainer#0-0-C-1] o.h.type.descriptor.sql.BasicExtractor   : extracted value ([id1_2_0_] : [BIGINT]) - [10]
2019-03-13 08:05:54.694 TRACE 4613 --- [ntainer#0-0-C-1] o.h.type.descriptor.sql.BasicExtractor   : extracted value ([labels2_2_0_] : [VARCHAR]) - [localdev]
2019-03-13 08:05:54.694 TRACE 4613 --- [ntainer#0-0-C-1] o.h.type.descriptor.sql.BasicExtractor   : extracted value ([labels_k3_0_] : [VARCHAR]) - [envoy.environment]
2019-03-13 08:05:54.694 TRACE 4613 --- [ntainer#0-0-C-1] o.h.type.descriptor.sql.BasicExtractor   : extracted value ([id1_2_0_] : [BIGINT]) - [10]
2019-03-13 08:05:54.694 TRACE 4613 --- [ntainer#0-0-C-1] o.h.type.descriptor.sql.BasicExtractor   : extracted value ([labels2_2_0_] : [VARCHAR]) - [MS90HCG8WL]
2019-03-13 08:05:54.694 TRACE 4613 --- [ntainer#0-0-C-1] o.h.type.descriptor.sql.BasicExtractor   : extracted value ([labels_k3_0_] : [VARCHAR]) - [envoy.hostname]
2019-03-13 08:05:54.694 TRACE 4613 --- [ntainer#0-0-C-1] o.h.type.descriptor.sql.BasicExtractor   : extracted value ([id1_2_0_] : [BIGINT]) - [10]
2019-03-13 08:05:54.694 TRACE 4613 --- [ntainer#0-0-C-1] o.h.type.descriptor.sql.BasicExtractor   : extracted value ([labels2_2_0_] : [VARCHAR]) - [DARWIN]
2019-03-13 08:05:54.694 TRACE 4613 --- [ntainer#0-0-C-1] o.h.type.descriptor.sql.BasicExtractor   : extracted value ([labels_k3_0_] : [VARCHAR]) - [envoy.os]
2019-03-13 08:05:54.695 DEBUG 4613 --- [ntainer#0-0-C-1] o.s.orm.jpa.JpaTransactionManager        : Initiating transaction commit
2019-03-13 08:05:54.695 DEBUG 4613 --- [ntainer#0-0-C-1] o.s.orm.jpa.JpaTransactionManager        : Committing JPA transaction on EntityManager [SessionImpl(1440863985<open>)]
2019-03-13 08:05:54.698 DEBUG 4613 --- [ntainer#0-0-C-1] o.s.orm.jpa.JpaTransactionManager        : Closing JPA EntityManager [SessionImpl(1440863985<open>)] after transaction
```

itzg:
@harageth I pushed a commit with a few more debug logs added to further confirm the transaction is closed before the resource event is published:

```
2019-03-13 08:34:42.870 DEBUG 10060 --- [ntainer#0-0-C-1] o.s.orm.jpa.JpaTransactionManager        : Closing JPA EntityManager [SessionImpl(57881245<open>)] after transaction
2019-03-13 08:34:42.870 DEBUG 10060 --- [ntainer#0-0-C-1] c.r.s.r.services.ResourceManagement      : Publishing resource event: ResourceEvent(resource=Resource(id=10, resourceId=development:0, labels={envoy.environment=localdev, envoy.hostname=MS90HCG8WL, envoy.os=DARWIN, envoy.arch=X86_64}, tenantId=aaaaaa, presenceMonitoringEnabled=true), oldLabels={envoy.os=DARWIN, envoy.arch=X86_64, envoy.environment=localdev, envoy.hostname=MS90HCG8WL}, operation=UPDATE, presenceMonitorChange=false)
```

src/test/java/com/rackspace/salus/resource_management/ResourceManagementTest.java
```diff
@@ -233,4 +283,122 @@ public void testRemoveResource() {
...
+        labels.put("region", "DFW");
+
+
+        ResourceCreate create = podamFactory.manufacturePojo(ResourceCreate.class);
+        create.setLabels(resourceLabels);
+        String tenantId = RandomStringUtils.randomAlphanumeric(10);
+        resourceManagement.createResource(tenantId, create);
+        entityManager.flush();
+
+        List<Resource> resources = resourceManagement.getResourcesFromLabels(labels, tenantId);
```
itzg: `labels` and `resourceLabels` seem to be flipped. The query here would be called when given a monitor find the resources, in which case the "subset scenario" is the typical case where the monitor has a subset of required labels to match against the larger set of the expected resource.

itzg: Please ignore my previous comment, I had gotten myself confused relative to the previous test case.

src/test/java/com/rackspace/salus/resource_management/ResourceManagementTest.java
```diff
@@ -233,4 +277,122 @@ public void testRemoveResource() {
...
+        assertEquals(1, resources.size()); //make sure we only returned the one value
```
itzg: This should match zero resources since it is impossible for a query with 4 labels to match all of its labels against a resource with only two labels.

harageth: But similar to MonitorManagement where we generally expect to match monitors with fewer labels to resources with more labels there will still be instances, albeit rare where we need to do the same here.

src/test/java/com/rackspace/salus/resource_management/ResourceManagementTest.java
```diff
@@ -233,4 +283,122 @@ public void testRemoveResource() {
         resource = resourceManagement.getResource(tenantId, create.getResourceId());
         assertThat(resource, nullValue());
     }
-}
\ No newline at end of file
+
+    @Test
+    public void testLabelMatchingQuery() {
```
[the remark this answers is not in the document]
harageth: *jedi mind trick* this test never existed

src/test/java/com/rackspace/salus/resource_management/ResourceManagementTest.java
```diff
@@ -233,4 +277,114 @@ public void testRemoveResource() {
...
+        resourceManagement.createResource(tenantId2, create);
+
+        List<Resource> resources = resourceManagement.getResourcesFromLabels(labels, tenantId);
+        assertEquals(1, resources.size()); //make sure we only returned the one value
+        assertEquals(tenantId, resources.get(0).getTenantId());
+        assertEquals(create.getResourceId(), resources.get(0).getResourceId());
+    }
+
+    @Test
+    public void testMatchResourceWithMultipleLabels() {
```
itzg: Please implement a negative test case where label count equals but one of the labels differ.

src/test/java/com/rackspace/salus/resource_management/ResourceManagementTest.java
```diff
@@ -233,4 +277,114 @@ public void testRemoveResource() {
...
+
+        List<Resource> resources = resourceManagement.getResourcesFromLabels(labels, tenantId);
+        assertEquals(1, resources.size()); //make sure we only returned the one value
+        assertEquals(tenantId, resources.get(0).getTenantId());
+        assertEquals(create.getResourceId(), resources.get(0).getResourceId());
+        assertEquals(labels, resources.get(0).getLabels());
+    }
+
+    @Test
+    public void testMatchResourceWithSupersetOfLabels() {
```
itzg: Please provide a negative test case, still where resource label count > query label count but one of the query labels has a different value, like env=staging

src/test/java/com/rackspace/salus/resource_management/ResourceManagementTest.java
itzg: Can you describe those instances in real world terms? 

In your unit test, architecture and region are effectively ignored, so the parameter description in the javadoc

https://github.com/racker/salus-telemetry-resource-management/pull/9/files#diff-d034cbc23c60da4b9dc850475e9f9768R386

needs to be reworded from:

> the labels that we need to match to

src/test/java/com/rackspace/salus/resource_management/ResourceManagementTest.java
```diff
@@ -233,4 +277,114 @@ public void testRemoveResource() {
...
+
+        List<Resource> resources = resourceManagement.getResourcesFromLabels(labels, tenantId);
+        assertEquals(1, resources.size()); //make sure we only returned the one value
+        assertEquals(tenantId, resources.get(0).getTenantId());
+        assertEquals(create.getResourceId(), resources.get(0).getResourceId());
+        assertEquals(resourceLabels, resources.get(0).getLabels());
+    }
+
+    @Test
+    public void testMatchResourceWithSubsetOfLabels() {
```
itzg: Please provide two alternate test cases where query label count is still greater than resource label count, but

1. Change one of the overlapping label values, such as env=prod
2. Change one of the non-overlapping label values, such as region=LON

src/test/java/com/rackspace/salus/resource_management/ResourceManagementTest.java
harageth: Sure. The situation we had talked about before of a brand new resource with only the labels that the system has assigned to it (I am not sure how many pre-setup labels we are creating but I can see: architecture, os, region, cloud). 
Then we have an already created monitor with labels: 
{arch:64, os: "linux", region:"IAD", cloud:"rackspace", env:"prod", redirects:false}

I think this would generally be the exception, but wanted to make sure I covered it still. 

Although now that I have written it out with a specific example I am now a little unsure. I am not sure people would want something configured for prod to run on a server that may or may not be a prod server.

itzg: It sounds like you now see what I'm getting at. I think your javadoc description of `labels` is correct and as you said, a user that wants a monitor running on env=prod would definitely want that assigned to a resource that explicitly is labelled with env=prod.

harageth: Furthermore it would have also caused an issue with duplicate monitors. I am imagining that there would be similar monitors in prod and test and if the resource doesn't have the environment listed then with what I suggested it would get all of the monitors for both prod and test. This seems like the worse scenario too. I will change this shortly and add the negative tests as well.

src/test/java/com/rackspace/salus/resource_management/ResourceManagementTest.java
```diff
@@ -233,4 +277,191 @@ public void testRemoveResource() {
...
+        String tenantId = RandomStringUtils.randomAlphanumeric(10);
+        resourceManagement.createResource(tenantId, create);
+        entityManager.flush();
+
+        List<Resource> resources = resourceManagement.getResourcesFromLabels(labels, tenantId);
+        assertEquals(0, resources.size());
+    }
+
+    @Test
+    public void testMatchResourceWithSubsetOfLabels() {
```
itzg: This now seems to be a duplicate of `testFailMatchResourceWithSubsetOfLabels`

src/test/java/com/rackspace/salus/resource_management/ResourceManagementTest.java
```diff
@@ -233,4 +277,191 @@ public void testRemoveResource() {
...
+
+    @Test
+    public void testFailedMatchResourceWithMultipleLabels() {
+        final Map<String, String> resourceLabels = new HashMap<>();
+        resourceLabels.put("os", "DARWIN");
+        resourceLabels.put("env", "test");
+
+        final Map<String, String> labels = new HashMap<>();
+        labels.put("os", "DARWIN");
+        labels.put("env", "prod"); //what happens if we dont supply labels?
```
itzg: Question in comment still needed since you now have `testSendNoLabels`?
> what happens if we dont supply labels?

src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
```diff
@@ -361,5 +381,112 @@ private void removePresenceMonitoring(String tenantId, String resourceId) {
...
+                            .setResourceId(resultSet.getString("resource_id"))
+                            .setTenantId(resultSet.getString("tenant_id"))
+                            .setPresenceMonitoringEnabled(resultSet.getBoolean("presence_monitoring_enabled"))
+                            .setLabels(theseLabels);
+                    prevId = resultSet.getLong("id");
+                    prevResource = r;
+                    resources.add(r);
+
+                }
+                iterate = resultSet.next();
```
itzg: Separately from this code and PR, it would be really good to confirm this `.next()` behavior here. The `ResultSet.next()` docs emphasize the cursor is initialized positioned before the first record, so this discrepancy still bothers me:

> A ResultSet cursor is initially positioned before the first row; the first call to the method next makes the first row the current row; the second call makes the second row the current row, and so on.

[From here](https://docs.oracle.com/javase/8/docs/api/java/sql/ResultSet.html#next--)

src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
```diff
@@ -361,5 +381,112 @@ private void removePresenceMonitoring(String tenantId, String resourceId) {
...
+            }
+        });
+
+        return resources;
+    }
+
+    public List<String> getResourceIdsWithEnvoyLabels(Map<String, String> labels, String tenantId) {
```
itzg: I had hacked in this method copying from what was `getResourcesFromLabels` at the time where the only difference was the final record processing. It now appears to have diverged from the query built in `getResourcesFromLabels` and should probably also be updated with the same inner selections.

itzg: Alternatively this method could be deleted to avoid having to maintain two methods with the same query. The downside is that the monitor management code will need to be updated to call the other REST API query.

src/test/java/com/rackspace/salus/resource_management/ResourceManagementTest.java
```diff
@@ -192,6 +208,34 @@ public void testNewEnvoyAttach() {
         assertThat(resource.getResourceId(), equalTo(attachEvent.getResourceId()));
     }
 
+    @Test
+    public void testEnvoyAttachAndQueryByLabels() {
```
itzg: This one also needs a negative test case where `labelsToQuery` is a superset of `labels`.

src/test/java/com/rackspace/salus/resource_management/ResourceManagementTest.java
harageth: No I dont think so. So long as you are fine with me throwing an IllegalArgumentException

src/main/java/com/rackspace/salus/resource_management/web/controller/ResourceApi.java
```diff
@@ -115,4 +130,16 @@ public void delete(@PathVariable String tenantId,
                        @PathVariable String resourceId) {
         resourceManagement.removeResource(tenantId, resourceId);
     }
+
+    @GetMapping("/tenant/{tenantId}/resourceLabels")
+    public List<Resource> getResourcesWithLabels(@PathVariable String tenantId,
+                                                 @RequestBody Map<String, String> labels) {
```
itzg:
```suggestion
                                                 @RequestParam Map<String, String> labels) {
```

I just noticed I had to make this change locally for the public API proxy to work without customizations -- GETs don't normally have a body.

src/test/java/com/rackspace/salus/resource_management/ResourceManagementTest.java
harageth: Good catch. I think we only need one of those two tests so I am keeping this one.

PR #10, opened 2019-03-05
Add query to get resourceIds for given Envoy labels

itzg:
# Resolves

For end to end integration https://jira.rax.io/browse/SALUS-205

# What

To keep things simple for now and focused on retrieving the resource IDs associated with attached Envoys, I added a query method and REST call focused on returning just the resource IDs given specifically Envoy advertised labels.

## How to test

Added a unit test to simulate an Envoy attach and then querying for its associated resource ID.

# TODO

I'll also push out a PR for the corresponding monitor management changes.

### DIFF
--- a/src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
+++ b/src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
@@ -19,35 +19,37 @@ package com.rackspace.salus.resource_management.services;
 import com.rackspace.salus.resource_management.web.model.ResourceCreate;
 import com.rackspace.salus.resource_management.web.model.ResourceUpdate;
 import com.rackspace.salus.telemetry.errors.ResourceAlreadyExists;
+import com.rackspace.salus.telemetry.messaging.AttachEvent;
+import com.rackspace.salus.telemetry.messaging.OperationType;
 import com.rackspace.salus.telemetry.messaging.ResourceEvent;
-import com.rackspace.salus.telemetry.model.*;
-import com.rackspace.salus.telemetry.messaging.*;
+import com.rackspace.salus.telemetry.model.NotFoundException;
+import com.rackspace.salus.telemetry.model.Resource;
+import com.rackspace.salus.telemetry.model.Resource_;
 import com.rackspace.salus.telemetry.repositories.ResourceRepository;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.stream.Stream;
+import javax.persistence.EntityManager;
+import javax.persistence.NoResultException;
+import javax.persistence.PersistenceContext;
+import javax.persistence.criteria.CriteriaBuilder;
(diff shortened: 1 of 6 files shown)


src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
```diff
@@ -405,5 +407,40 @@ private void removePresenceMonitoring(String tenantId, String resourceId) {
...
+        int i = 0;
+        envoyLabels.size();
+        for (Map.Entry<String, String> entry : envoyLabels.entrySet()) {
+            if (i > 0) {
+                builder.append(" OR ");
+            }
+            builder.append("(labels = ? AND labels_key = ?)");
+            i++;
+        }
+        builder.append(" GROUP BY id HAVING COUNT(id) = ?)");
```
harageth: We should add the ORDER BY at the end of this. 

You can also use Named Parameters. Might make it easier to read what is happening.

src/main/resources/application-dev.yml
```diff
@@ -3,7 +3,7 @@ server:
 spring:
   jpa:
     hibernate:
-      ddl-auto: create
+      ddl-auto: update
```
harageth: What is happening here? between the create and update? I thought flyway was handling database versioning for us?

itzg: We're not using flyway yet :) There's an open story for that.

I just didn't want hibernate blowing away the tables in between restarts. Notice this is the **dev** config file, so has no relationship to schema management in production.

harageth: I was hoping that was what you were going to say. I am a little tired of running the same API calls against the server every time I have to restart my local server.

src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
itzg: That operation didn't support named parameters :( So I had to go old school with only positional.

Now that you mention `ORDER BY` I'm worried this will be returning duplicates of the resource IDs. Does it need an outer `GROUP BY` too?

harageth: Hmmmm let me look more closely. The Group By should only be needed in the inner query that is filtering on labels.

harageth: The query looks good as is. Especially since you are only returning a list of id's. 
The `GROUP BY` is really only needed in the inner query that is matching the labels because we need to group on the specific resources to get the `COUNT` to match to the number of labels we need to match against.  Its probably the most subtle part of the query. It's the part that we could remove to alter the query from "MATCH ALL" to "MATCH SOME" if we ever wanted to do that sort of search.

itzg: Do it doesn't need the `ORDER BY`?

harageth: It's still good practice to always `ORDER BY` but since you are only returning id's it doesn't really matter. You could drive the point home by returning it in an object that by its nature can't be ordered in a meaningful sense.

PR #11, opened 2019-03-06
Added cloudbuild files

dgvigil:
# Resolves
Jira story/task
* [SALUS-233](https://jira.rax.io/browse/SALUS-233) 
* [SALUS-246](https://jira.rax.io/browse/SALUS-246) 

# What
Added cloudbuild.yaml and cloudbuild-deploy.yaml for automated tests and deployments

### DIFF
--- a/cloudbuild-deploy.yaml
+++ b/cloudbuild-deploy.yaml
@@ -1,0 +1,92 @@
+substitutions:
+  _GCS_CACHE_BUCKET: salus-cache
+  _SALUS_PROJECT: salus-telemetry-resource-management
+
+steps:
+
+    # This is ugly because of
+    # https://github.com/GoogleContainerTools/jib/issues/1500#issuecomment-466207421
+  - id: FIX_DOCKER
+    name: gcr.io/cloud-builders/mvn
+    waitFor: ['-']
+    dir: /root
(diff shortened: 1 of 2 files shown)


cloudbuild-deploy.yaml
```diff
@@ -0,0 +1,92 @@
...
+
+  - id: COMPILE_AND_PUSH_CONTAINER
+    name: 'gcr.io/cloud-builders/mvn'
+    env:
+    - 'SHORT_SHA=$SHORT_SHA'
+    args:
+    - compile
+    # Runs the Jib build by using the latest version of the plugin.
+    # To use a specific version, configure the plugin in the pom.xml.
+    - jib:build
```
jjbuchan: @itzg guessing we need to add a jib `configuration` block to the pom?

itzg: The plugin is already enabled

https://github.com/racker/salus-telemetry-resource-management/blob/master/pom.xml#L111

and configured in the parent pom

https://github.com/racker/salus-app-base/blob/master/pom.xml#L89

jjbuchan: ok, didnt know about the parent pom part

PR #12, opened 2019-03-16
fix api test

GeorgeJahad:
The test was broken, this fixes a bean creation error.


harageth:
Ran the tests before and after to confirm that it works. I am not sure what caused the need for this change but I can confirm it works.

GeorgeJahad:
This is a bug that was introduced when I added the monitorRepository to the model.  I had a fixed branch  for it then, but I forgot to create a PR from that branch, (because I was working on the monitor manager, not the resource manager.)   Another example of not pushing all of the changes in all the git submodules.

PR #13, opened 2019-03-22
Envoy now does namespacing of Envoy discovered labels

itzg:
# Resolves

Part of https://jira.rax.io/browse/SALUS-265

# What

As of https://github.com/racker/salus-telemetry-envoy/pull/34 the Envoy is doing its own namespacing of the labels it discovers, so that no longer needs to be done in the resource manager. It does, however, still need to enforce that a resource update operation does not include user labels that conflict with the system namespaces.

## How to test

Unit test updated

### DIFF
--- a/src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
+++ b/src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
@@ -19,18 +19,20 @@ package com.rackspace.salus.resource_management.services;
 import com.rackspace.salus.resource_management.web.model.ResourceCreate;
 import com.rackspace.salus.resource_management.web.model.ResourceUpdate;
 import com.rackspace.salus.telemetry.errors.ResourceAlreadyExists;
+import com.rackspace.salus.telemetry.messaging.AttachEvent;
+import com.rackspace.salus.telemetry.messaging.OperationType;
 import com.rackspace.salus.telemetry.messaging.ResourceEvent;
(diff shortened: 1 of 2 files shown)


src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
```diff
@@ -217,6 +221,16 @@ public Resource updateResource(String tenantId, String resourceId, @Valid Resour
                     != updatedValues.getPresenceMonitoringEnabled().booleanValue();
         }
 
+        for (Entry<String, String> labelEntry : updatedValues.getLabels().entrySet()) {
+            final String labelName = labelEntry.getKey();
+            if (!LabelNamespaces.validateUserLabel(labelName)) {
+                throw new IllegalArgumentException(String
```
harageth: Should this be an IllegalArgumentException? Its valid. I am fine if you wanna keep it like this because I can't think of a built in exception that would match better. Perhaps InvalidParameterException? It just doesn't seem quite right, especially if we display it to the user in a similar format as other IllegalArgumentExceptions.

jjbuchan:
Somehow there's conflicts here now.

src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
itzg: Yep, it should be. I want that to pop out as a 400 for the REST API user since in this case they would have attempted to update a resource with a disallowed label.

PR #14, opened 2019-03-25
Remove envoy label namespace

harageth:
# Resolves

    https://jira.rax.io/browse/SALUS-266

# What

    We decided we didn't want to have label namespaces built into the system and make the user put those in when he sends the labels to query for

# How

    It just rips out the portions that added those namespaces

## How to test

    Run the ResourceManagementTest

# Why

    Not many other ways to remove something other than removing it.

### DIFF
--- a/src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
+++ b/src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
@@ -62,8 +62,6 @@ public class ResourceManagement {
     @PersistenceContext
     private final EntityManager entityManager;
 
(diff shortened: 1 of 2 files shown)


itzg:
@harageth I'm so sorry, I should have remembered the Jira issue I worked intersected with the Jira queued up for you. I had already made these types of changes in https://github.com/racker/salus-telemetry-resource-management/pull/13 , which is why the code started to look really familiar as I reviewed it.

harageth:
@itzg I was going through it and was like "huh I wonder if Geoff had done this with his PR thats open"

Its all good though it wasn't like it was that much work.

PR #16, opened 2019-04-09
update resource manager for new resource event

GeorgeJahad:
# Resolves

https://jira.rax.io/browse/SALUS-293
depends on: https://github.com/racker/salus-telemetry-model/pull/40

# How

removes the setting of the extra fields that are no longer used by the resource event, when resourceMan creates it.

## How to test

unit tests


jjbuchan:
eh... except from the conflicts.  I'll fix them.

itzg:
@jjbuchan yep, still looks good to me

jjbuchan:
It was a little too painful to rebase this with all the label changes that happened.  Instead I just started from scratched and made the same changes George did, and then put in a few extra tweaks.

PR #17, opened 2019-04-12
Fix updateEnvoyLabels and remove getResourceIdsWithEnvoyLabels

itzg:
# Resolves

Part of https://jira.rax.io/browse/SALUS-305

# What

We had previously identified that `getResourceIdsWithEnvoyLabels` was an awkward method to maintain since it mostly duplicated the other labels query. Since I was already making changes in monitor management, I went ahead and shifted the usage of this "get IDs" variant over to the main "get resources" query.

During testing I also found that `updateEnvoyLabels` wasn't properly handling the case when new envoy labels were added during an update. That also led me to find it needed to leverage the label namespacing to differentiate labels provided via the resource API from those provided during envoy attach.

## How to test

Unit tests were updated

### DIFF
--- a/src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
+++ b/src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
@@ -37,6 +37,7 @@ import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import javax.persistence.EntityManager;
 import javax.persistence.NoResultException;
@@ -218,6 +219,7 @@ public class ResourceManagement {
                 .setTenantId(tenantId)
                 .setResourceId(newResource.getResourceId())
                 .setLabels(newResource.getLabels())
+                .setMetadata(newResource.getMetadata())
                 .setPresenceMonitoringEnabled(newResource.getPresenceMonitoringEnabled())
                 .setRegion(newResource.getRegion());
 
@@ -245,14 +247,22 @@ public class ResourceManagement {
             presenceMonitoringStateChange = resource.getPresenceMonitoringEnabled().booleanValue()
                     != updatedValues.getPresenceMonitoringEnabled().booleanValue();
         }
+
         if (updatedValues.getLabels() != null) {
             checkLabels(updatedValues.getLabels());
+
+            final Map<String, String> mergedLabels = Stream
+                .concat(
+                    updatedValues.getLabels().entrySet().stream(),
+                    oldLabels.entrySet().stream()
+                        .filter(entry -> labelHasNamespace(entry.getKey(), LabelNamespaces.AGENT))
+                )
+                .collect(Collectors.toMap(Entry::getKey, Entry::getValue));
+
+            resource.setLabels(mergedLabels);
         }
 
         PropertyMapper map = PropertyMapper.get();
-        map.from(updatedValues.getLabels())
-                .whenNonNull()
-                .to(resource::setLabels);
         map.from(updatedValues.getMetadata())
                 .whenNonNull()
                 .to(resource::setMetadata);
@@ -434,7 +444,7 @@ public class ResourceManagement {
 
         MapSqlParameterSource paramSource = new MapSqlParameterSource();
         paramSource.addValue("tenantId", tenantId);//AS r JOIN resource_labels AS rl
-        StringBuilder builder = new StringBuilder("SELECT * FROM resources JOIN resource_labels AS rl WHERE resources.id = rl.id AND resources.id IN ");
+        StringBuilder builder = new StringBuilder("SELECT resources.id as id FROM resources JOIN resource_labels AS rl WHERE resources.id = rl.id AND resources.id IN ");
         builder.append("(SELECT id from resource_labels WHERE id IN ( SELECT id FROM resources WHERE tenant_id = :tenantId) AND resources.id IN ");
         builder.append(" (SELECT id FROM resource_labels WHERE ");
         int i = 0;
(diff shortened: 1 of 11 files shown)


src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
```diff
@@ -324,24 +329,43 @@ public void handleEnvoyAttach(AttachEvent attachEvent) {
...
+            if (envoyLabelValue != null) {
+                if (!envoyLabelValue.equals(value)) {
+                    // goal 2
                     updated.set(true);
-                    resourceLabels.put(key, value);
+                    resourceLabels.put(key, envoyLabelValue);
                 }
-            } else {
+                newEnvoyLabelKeys.remove(key);
+            } else if (labelHasNamespace(key, LabelNamespaces.AGENT)) { // goal 1
```
harageth: This is only true if we have Labels that match the namespace. But we are then removing those labels from the resource. Which means that we can never match namespaced labels from the Monitor to the Resource.. Unless I am missing something.

itzg: I'm confused about what you mean, but then the original (and new) code was confusing too. Please refer to the new unit tests:

- https://github.com/racker/salus-telemetry-resource-management/pull/17/files#diff-0a75ec9c8c786011e786f8125a9d8eafR285
- https://github.com/racker/salus-telemetry-resource-management/pull/17/files#diff-0a75ec9c8c786011e786f8125a9d8eafR324

and let me know if any other unit tests are needed there.

src/test/java/com/rackspace/salus/resource_management/ResourceManagementTest.java
```diff
@@ -258,13 +274,102 @@ public void testEnvoyAttachAndFailedQueryByLabels() {
...
+
+        final Optional<Resource> actualResource = resourceRepository.findById(resource.getId());
+
+        assertThat(actualResource.isPresent(), equalTo(true));
+        //noinspection OptionalGetWithoutIsPresent
+        assertThat(actualResource.get().getLabels(), equalTo(envoyLabels));
+
+        verify(kafkaEgress).sendResourceEvent(resourceEventArg.capture());
+        assertThat(resourceEventArg.getValue().getResource().getResourceId(), equalTo("r-1"));
+        assertThat(resourceEventArg.getValue().getResource().getLabels(), equalTo(envoyLabels));
```
harageth: Ok I think I need to re-read the code because it looked to me like it was removing the envoy labels.

itzg: That outer `forEach` is iterating over the existing resource labels and then the `else if` only applies when the newly attached envoy labels don't include the current key of the iteration.

src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
harageth: I am tagging this as one of the things I want to talk over on Monday

src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
```diff
@@ -324,24 +329,43 @@ public void handleEnvoyAttach(AttachEvent attachEvent) {
      * @param envoyLabels The list of labels received from a newly connected envoy.
      */
     private void updateEnvoyLabels(Resource resource, Map<String, String> envoyLabels) {
```
itzg: As a note to myself, the `envoyLabels` here have already been namespace'd over here

https://github.com/racker/salus-telemetry-ambassador/blob/ffad5023764abfa22dc6a33252349ce29023cb4e/src/main/java/com/rackspace/salus/telemetry/ambassador/services/EnvoyRegistry.java#L160

src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
```diff
@@ -422,68 +461,16 @@ private void removePresenceMonitoring(String tenantId, String resourceId) {
...
-                    Resource r = new Resource()
-                            .setId(resultSet.getLong("id"))
-                            .setResourceId(resultSet.getString("resource_id"))
-                            .setTenantId(resultSet.getString("tenant_id"))
-                            .setPresenceMonitoringEnabled(resultSet.getBoolean("presence_monitoring_enabled"))
-                            .setLabels(theseLabels);
-                    prevId = resultSet.getLong("id");
-                    prevResource = r;
-                    resources.add(r);
+        final List<Long> resourceIds = namedParameterTemplate.query(builder.toString(), paramSource,
```
itzg: FYI I swapped out the result processing into two distinct steps because

1. With the introduction of metadata and given that each query result row is repeated by the number of labels present, the overall size of the SQL response body was going to grow unbounded.
2. Deserializing each of the metadata fields was going to get tedious with explicit use of ObjectMapper and then following the combination logic that was there.
3. The use of Spring Data's `findAllById` achieves a fair compromise where the overall method is still limited to two trips to the DB, keeping both at their bare minimum response size.
4. Using the repository for response deserialization keeps the overall method focused on the complexity of the query and keeps the simple parts simple.

src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
```diff
@@ -451,37 +461,14 @@ private void removePresenceMonitoring(String tenantId, String resourceId) {
...
+            (resultSet, rowIndex) -> resultSet.getLong(1)
```
harageth: I would like to see this referenced as the parameterized column name.

harageth: ohhh nvm I see whats going on now. I am fine with this being like this or as the parameterized column

src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
harageth: I like this compromise. It maintains the logic with minimal impact and allows us to deserialize in one location rather than two.

PR #19, opened 2019-04-24
Add ResourceApiClient

itzg:
# Resolves

Part of https://jira.rax.io/browse/SALUS-311

# What

Adds an internal REST API client similar to the [MonitorApiClient](https://github.com/racker/salus-telemetry-monitor-management/blob/master/src/main/java/com/rackspace/salus/monitor_management/web/client/MonitorApiClient.java)

# How

So far includes the query method needed by the monitor management when processing a resource event.

## How to test

New unit tests have been added.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,28 @@ kafka-console-consumer --bootstrap-server localhost:9093 --topic telemetry.resou
 
 You can trigger these events to be posted by utilizing some of the API operations below.
 
+# Resource Management REST API Client
+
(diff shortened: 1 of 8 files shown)


itzg:
I'll wait just a little longer to merge this PR in case I find more changes needed during monitor management work.

harageth:
Seems reasonable. Will we see a PR shortly for monitor management that starts using it?

itzg:
> Seems reasonable. Will we see a PR shortly for monitor management that starts using it?

Yes, hopefully in a day or two here.

PR #20, opened 2019-04-25
Swagger documentation

harageth:
# Resolves

    SALUS-318

# What

Add documentation for our API endpoints

# How

We need to generate the documentation on the internal API's (resource management, monitor management, etc...) but they have slightly different paths compared to the public API.

## How to test

pull this branch and run the Compile Maven command. (you need to compile so that the annotations get compiled into the source for picked by the maven swagger plugin). You should see output in {basedir}/target/generated/swagger/swagger.json

There is a new maven endpoint under plugins->exec->java
This new target will run the converter on the swagger json and output to {basedir}/target/generated/swagger/convertedOutput.json
This is how to generate the swagger for the public api

# Why

We can't really run the documentation processes on the public api because it just proxies all of the data through. So we basically need to create the documentation on the internal services and alter the path to reflect the public api path.

# TODO

Add a function to the mustache html template that will take the paths and create the correct path


pom.xml
```diff
@@ -99,6 +99,27 @@
...
+      <artifactId>swagger-core</artifactId>
+      <version>1.5.22</version>
+      <exclusions>
+        <exclusion>
+          <groupId>javax.ws.rs</groupId>
+          <artifactId>jsr311-api</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
+    <dependency>
```
itzg: I think the dependency within the plugin 

https://github.com/racker/salus-telemetry-resource-management/pull/20/files#diff-600376dffeb79835ede4a0b285078036R197

was the actual one needed.

pom.xml
```diff
@@ -130,6 +151,76 @@
...
+                <scheme> http </scheme>
+              </schemes>
+              <info>
+                <title>Salus Resource Management</title>
+                <version>0.1.0</version>
+              </info>
+              <outputPath>${project.build.directory}/generated/swagger/document.html</outputPath>
+              <swaggerDirectory>${project.build.directory}/generated/swagger</swaggerDirectory>
+              <outputFormats>json</outputFormats>
+ <!--             <securityDefinitions>
```
itzg: Uncomment or delete

pom.xml
```diff
@@ -130,6 +151,76 @@
...
+        <version>1.2.1</version>
+        <executions>
+          <execution>
+            <goals>
+              <goal>java</goal>
+            </goals>
+          </execution>
+        </executions>
+        <configuration>
+          <mainClass>com.rackspace.salus.resource_management.swaggerJsonConverter</mainClass>
```
itzg: Class names should start with uppercase, but it looks specifically like the filename needs to start with an uppercase.

You'll now need to set this

https://docs.spring.io/spring-boot/docs/2.1.4.RELEASE/maven-plugin/repackage-mojo.html#mainClass

in the spring boot plugin's config since there's now two main methods.

Alternatively you could put the converter under `src/test/java` and then tell the exec to use `test` scope here

https://www.mojohaus.org/exec-maven-plugin/java-mojo.html#classpathScope

This is actually a preferable approach since it avoids "polluting" the released jar with build-time code.

src/main/java/com/rackspace/salus/resource_management/swaggerJsonConverter.java
```diff
@@ -0,0 +1,4 @@
+package com.rackspace.salus.resource_management;
+
+public class SwaggerJsonConverter {
```
itzg: The filename should start with an uppercase to match case of the class. I'm surprised the compiler allows this.

itzg: Are there more changes to push? There's no `main` method (or anything) in here.

src/main/java/com/rackspace/salus/resource_management/web/controller/ResourceApiController.java
```diff
@@ -45,9 +47,19 @@
 import org.springframework.web.bind.annotation.ResponseStatus;
 import org.springframework.web.bind.annotation.RestController;
 import org.springframework.web.servlet.mvc.method.annotation.SseEmitter;
+import io.swagger.annotations.*;
+
 
 @Slf4j
 @RestController
+@Api(value = "/api", description = "Resource operations", authorizations = {
```
itzg: Does the `value` default to the `@RequestMapping` path if left off? If so, it would be good to not duplicate the path when not necessary.

src/main/java/com/rackspace/salus/resource_management/web/controller/ResourceApiController.java
```diff
@@ -45,9 +47,19 @@
...
+    @Authorization(value = "repose_auth",
+        scopes = {
+            @AuthorizationScope(scope = "write:resource", description = "modify resources in your account"),
```
itzg: Technically we don't have this level of RBAC yet, but I guess it doesn't hurt to declare what it probably will be.

templates/markdown.hbs
```diff
@@ -0,0 +1,108 @@
...
+
+{{description}}
+
+{{#contact}}
+[**Contact the developer**](mailto:{{email}})
+{{/contact}}
+
+**Version** {{version}}
+
+[**Terms of Service**]({{termsOfService}})
```
itzg: There isn't a TOS link declared, so can probably remove this from template for now.

src/main/java/com/rackspace/salus/resource_management/swaggerJsonConverter.java
harageth: Huh I apparently had an issue with something when I tried to refactor this file to have the first character be upper case.

pom.xml
```diff
@@ -130,6 +151,76 @@
...
+        </dependencies>
+      </plugin>
+
+      <plugin>
+        <groupId>org.codehaus.mojo</groupId>
+        <artifactId>exec-maven-plugin</artifactId>
+        <version>1.2.1</version>
+        <executions>
+          <execution>
+            <goals>
```
itzg: If you insert a `<phase>compile</phase>` here, it'll include this execution during the regular build. Since maven runs plugins in declared order within a phase, it'll guarantee this runs after the swagger.json is generated in plugin execution above.

src/main/java/com/rackspace/salus/resource_management/web/controller/ResourceApiController.java
harageth: It looks like it does. In all of the other annotations that `value` is required so I just assumed it was here too

pom.xml
harageth: That seems reasonable. It will almost immediately move again though because it will need to be more widely available when I add documentation for the other services

pom.xml
```diff
@@ -99,17 +99,6 @@
       <version>7.2.1.RELEASE</version>
       <scope>test</scope>
     </dependency>
-    <dependency>
```
itzg: I think you deleted the wrong dependency. I meant for you to delete `swagger-hibernate-validations`. `swagger-core` will be needed at compile time for the `@Api` type annotations.

src/main/java/com/rackspace/salus/resource_management/SwaggerJsonConverter.java
```diff
@@ -0,0 +1,60 @@
...
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.Map;
+import java.util.Scanner;
+
+public class SwaggerJsonConverter {
+
+    public static void main(String[] args) throws Exception {
+        ObjectMapper mapper = new ObjectMapper();
+        String content = new Scanner(new File(args[0]+"/swagger.json")).useDelimiter("\\Z").next();
```
itzg: What's delimiter "\\Z"?

src/main/java/com/rackspace/salus/resource_management/SwaggerJsonConverter.java
```diff
@@ -0,0 +1,60 @@
...
+        ObjectNode root = (ObjectNode)mapper.readTree(content);
+        Map<String, JsonNode> temp = new HashMap();
+        for (Iterator<Map.Entry<String, JsonNode>> it = root.get("paths").fields(); it.hasNext(); ) {
+            Map.Entry<String, JsonNode> elt = it.next();
+            if (elt.getKey().contains("tenant"))
+            {
+                String newKey = elt.getKey().replace("tenant/{tenantId}/", "");
```
itzg: I think you need to use "resources/" as the replacement text to arrive at the public API path like:

https://github.com/racker/salus-telemetry-api/blob/master/public/src/main/java/com/rackspace/salus/telemetry/api/web/ResourcesController.java#L88

src/main/java/com/rackspace/salus/resource_management/SwaggerJsonConverter.java
```diff
@@ -0,0 +1,60 @@
...
+
+            } else {
+                it.remove();
+            }
+        }
+        ObjectNode pathNode = mapper.getNodeFactory().objectNode();
+
+        temp.forEach((key, node)->{
+            pathNode.set(key, node);
+            //for some reason this works but placing it in the path doesnt...
```
itzg: That's what I would expect since concurrent modification of a collection being iterated isn't generally supported. `Iterator.remove()` is a special case where the collection's iterator can coordinate the concurrent change while iterating since it has positioned past the removed entry.

You can remove this comment now.

src/test/java/com/rackspace/salus/resource_management/Converter/SwaggerJsonConverter.java
```diff
@@ -0,0 +1,59 @@
+package com.rackspace.salus.resource_management.Converter;
```
itzg: Change `Converter` to `converter` to be a proper package name.

src/test/java/com/rackspace/salus/resource_management/Converter/SwaggerJsonConverter.java
```diff
@@ -0,0 +1,59 @@
...
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.Map;
+import java.util.Scanner;
+
+public class SwaggerJsonConverter {
+
+    public static void main(String[] args) throws Exception {
+        ObjectMapper mapper = new ObjectMapper();
+        String content = new Scanner(new File(args[0]+"/swagger.json")).useDelimiter("\\Z").next();
```
itzg: What is delimiter "\\Z"?

src/test/java/com/rackspace/salus/resource_management/Converter/SwaggerJsonConverter.java
```diff
@@ -0,0 +1,59 @@
...
+        ObjectNode root = (ObjectNode)mapper.readTree(content);
+        Map<String, JsonNode> temp = new HashMap();
+        for (Iterator<Map.Entry<String, JsonNode>> it = root.get("paths").fields(); it.hasNext(); ) {
+            Map.Entry<String, JsonNode> elt = it.next();
+            if (elt.getKey().contains("tenant"))
+            {
+                String newKey = elt.getKey().replace("tenant/{tenantId}/", "");
```
itzg: I think the replace text needs to be "resources/" to align with the public API paths such as

https://github.com/racker/salus-telemetry-api/blob/master/public/src/main/java/com/rackspace/salus/telemetry/api/web/ResourcesController.java#L88

src/test/java/com/rackspace/salus/resource_management/Converter/SwaggerJsonConverter.java
```diff
@@ -0,0 +1,59 @@
...
+
+            } else {
+                it.remove();
+            }
+        }
+        ObjectNode pathNode = mapper.getNodeFactory().objectNode();
+
+        temp.forEach((key, node)->{
+            pathNode.set(key, node);
+            //for some reason this works but placing it in the path doesnt...
```
itzg: That's expecteds since concurrent modification of a collection while iterating is generally not allowed. `Iterator.remove()` is a special case since the iterator has knowledge to skip past the change.

You can remove your comment.

src/test/java/com/rackspace/salus/resource_management/Converter/SwaggerJsonConverter.java
harageth: They all have resources after `/tenant/{tenantId}` in their path already. Except for the label matching function. If I did what you were suggesting then it would double the `resources/` in the path. And for the resourceLabels function it matches what is in the public api path since it doesn't have `resources/` in the path at all.

src/test/java/com/rackspace/salus/resource_management/Converter/SwaggerJsonConverter.java
harageth: Yeah I forgot to. This comment is why I did create the pathNode above though.

src/test/java/com/rackspace/salus/resource_management/Converter/SwaggerJsonConverter.java
harageth: https://stackoverflow.com/questions/22350037/behavior-of-using-z-vs-z-as-scanner-delimiter

end of input except for the final terminator, if any

harageth: Basically... Read the whole file in

pom.xml
```diff
@@ -99,6 +99,16 @@
       <version>7.2.1.RELEASE</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
```
itzg: I pulled down your branch and replacing these two with just

```
    <dependency>
      <groupId>io.swagger</groupId>
      <artifactId>swagger-core</artifactId>
      <version>1.5.6</version>
    </dependency>
```

worked for me. The annotations actually come from `swagger-annotations`, so even just that would be fine.

image

pom.xml
```diff
@@ -130,6 +140,72 @@
...
+      </plugin>
+
+      <plugin>
+        <groupId>org.codehaus.mojo</groupId>
+        <artifactId>exec-maven-plugin</artifactId>
+        <version>1.2.1</version>
+
+        <executions>
+          <execution>
+            <phase>compile</phase>
```
itzg: While trying out your branch I realized this now needs to be `test-compile` phase since the utility is in test classpath now.

PR #22, opened 2019-04-30
Add getExpectedEnvoys to API interface

jjbuchan:
# What

Adds a new endpoint to the resource api client.

# How

I added it to the interface and then implemented a version of it in both the controller and the client.  It is unlikely the new controller method will be used, as the clients will want to get a streaming result vs. a large bulk payload, however, it had to be implemented since it is defined in the interface.

Implemented a likely unnecessary method seemed like the best option.  The alternatives are defining a default method in the interface that does nothing or moving away from using the interface. 

## How to test

Run tests.

# Why

Presence monitoring now utilizes this api client, and a method for this endpoint was required.

### DIFF
--- a/src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
+++ b/src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
@@ -144,8 +144,8 @@ public class ResourceManagement {
         return new PageImpl<>(resources, page, resources.size());
     }
 
-    public List<Resource> getAllTenantResources(String tenantid) {
-        return resourceRepository.findAllByTenantId(tenantid);
+    public List<Resource> getAllTenantResources(String tenantId) {
+        return resourceRepository.findAllByTenantId(tenantId);
     }
 
     /**
@@ -175,14 +175,7 @@ public class ResourceManagement {
      * @return List of resources.
      */
     public List<Resource> getExpectedEnvoys() {
-        CriteriaBuilder cb = entityManager.getCriteriaBuilder();
-        CriteriaQuery<Resource> cr = cb.createQuery(Resource.class);
-        Root<Resource> root = cr.from(Resource.class);
-
-        cr.select(root).where(
-                cb.equal(root.get(Resource_.presenceMonitoringEnabled), true));
-
-        return entityManager.createQuery(cr).getResultList();
+        return resourceRepository.findAllByPresenceMonitoringEnabled(true);
     }
 
     /**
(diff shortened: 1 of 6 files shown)


src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
```diff
@@ -169,6 +169,22 @@ public Resource getResource(String tenantId, String resourceId) {
         return entityManager.createQuery(cr).getResultStream();
     }
 
+    /**
+     * Get a list of all resources where the presence monitoring is enabled.
+     *
+     * @return List of resources.
+     */
+    public List<Resource> getExpectedEnvoys() {
+        CriteriaBuilder cb = entityManager.getCriteriaBuilder();
```
itzg: Given our conversation yesterday, I would prefer this be declared as a method of `ResourceRepository`:

```
  List<Resource> findAllByPresenceMonitoringEnabled(boolean value);
```

or even

```
  List<Resource> findAllByPresenceMonitoringEnabledIsTrue();
```


and let Spring Data manage the query details.

src/test/java/com/rackspace/salus/resource_management/web/client/ResourceApiClientTest.java
```diff
@@ -52,21 +53,23 @@
   public static class ExtraTestConfig {
     @Bean
     public ResourceApiClient resourceApiClient(RestTemplateBuilder restTemplateBuilder) {
-      return new ResourceApiClient(restTemplateBuilder.build());
+      return new ResourceApiClient(objectMapper, restTemplateBuilder.build());
```
itzg: So, Spring autowired the static `objectMapper` prior to the bean factory being called? That's very handy to know.

itzg:
I'm still not totally sure we're using SSE for the correct purpose, so that led me to discover there is indeed support for serializing Java8 streams:

https://fasterxml.github.io/jackson-modules-java8/javadoc/datatypes/2.9/com/fasterxml/jackson/datatype/jdk8/StreamSerializer.html

however, it's not clear how you deserialize those. 

Overall, let's stick with SSE, but just wanted to point this out since we talked about streams the other day @jjbuchan .

src/test/java/com/rackspace/salus/resource_management/web/client/ResourceApiClientTest.java
jjbuchan: I had meant tot remove the `@Autowired` from object mapper below.  It is no longer needed.  I had to create the OM myself vs. autowiring, otherwise this would fail.

jjbuchan:
Made a couple minor changes.  I'm going to look into using a stream instead of SSE in another PR... I think I have a lead there.

PR #23, opened 2019-04-30
WIP: Use streams instead of SseEmitter for expected envoy request

jjbuchan:
# What

This doesn't work yet, but rather than throwing it away, I'll keep it here for now in case we stumble upon something that makes it work with Spring MVC.

Rather than using the SseEmitter for our (currently) one streaming api response, we could use some more native/clean options.

# How

This attempts to switch over to just using a `Stream` object, but as suspected it doesn't actually stream the response, it gets passed through as an array of resources.

## How to test

Run tests and manually verify the response output in console.

`ResourceApi` should be improved as it will currently passes if you force the content type with `@GetMapping(value = "/envoys", produces = "application/stream+json")` due to using `stringContainsInOrder` and the current expected string is wrapped in `[ ]`.

# Why
So things are "nicer".  SseEmitter includes junk data in the response that we have to strip out.  Currently it also works differently than any other endpoint.  It would be better if things were more consistent across all.

This could be accomplished with WebFlux, but it's not worth switching over to that for this one thing.

# TODO

If this ever works, we'd have to redeploy the other services that use this api client.

### DIFF
--- a/src/main/java/com/rackspace/salus/resource_management/repositories/ResourceRepository.java
+++ b/src/main/java/com/rackspace/salus/resource_management/repositories/ResourceRepository.java
@@ -18,6 +18,8 @@ package com.rackspace.salus.resource_management.repositories;
 
 import com.rackspace.salus.telemetry.model.Resource;
 import java.util.List;
+import java.util.stream.Stream;
+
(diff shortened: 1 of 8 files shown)


jjbuchan:
Closing, but it's here if we want to go back to it.

PR #26, opened 2019-05-02
Remove CriteriaBuilder and use repository methods instead

jjbuchan:
# What

Removes all the `CriteriaBuilder` logic for building sql queries and replaces it by using standard repository options.

# How

The "how" is fairly obvious.  The only real decision was to start using `Optional` fields where we are getting at most one returned object.  I like this overall, but the downside is it makes the tests a little more tedious due to having to do various `isPresent()` and `get()` calls.

## How to test

Run all tests.  They still work!

# Why

Cleans up the code and standardizes things across all repos.

# TODO

Incoming PRs for monitor management and model repo... I think that'll be it.

### DIFF
--- a/src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
+++ b/src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
@@ -84,6 +84,16 @@ public class ResourceManagement {
     }
 
     /**
+     * Tests whether the resource exists on the given tenant.
+     * @param tenantId The tenant owning the resource.
+     * @param resourceId The unique value representing the resource.
+     * @return True if the resource exists on the tenant, otherwise false.
+     */
+    public boolean exists(String tenantId, String resourceId) {
+        return resourceRepository.existsByTenantIdAndResourceId(tenantId, resourceId);
+    }
+
+    /**
      * Gets an individual resource object by the public facing id.
      * @param tenantId The tenant owning the resource.
      * @param resourceId The unique value representing the resource.
@@ -157,8 +167,7 @@ public class ResourceManagement {
      * @throws ResourceAlreadyExists
      */
     public Resource createResource(String tenantId, @Valid ResourceCreate newResource) throws IllegalArgumentException, ResourceAlreadyExists {
-        Optional<Resource> existing = getResource(tenantId, newResource.getResourceId());
-        if (existing.isPresent()) {
+        if (exists(tenantId, newResource.getResourceId())) {
             throw new ResourceAlreadyExists(String.format("Resource already exists with identifier %s on tenant %s",
                     newResource.getResourceId(), tenantId));
         }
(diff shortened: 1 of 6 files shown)


src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
```diff
@@ -133,40 +109,19 @@ public Resource getResource(String tenantId, String resourceId) {
...
     public Stream<Resource> getResources(boolean presenceMonitoringEnabled) {
-        CriteriaBuilder cb = entityManager.getCriteriaBuilder();
-        CriteriaQuery<Resource> cr = cb.createQuery(Resource.class);
-        Root<Resource> root = cr.from(Resource.class);
-
-        cr.select(root).where(
-                cb.equal(root.get(Resource_.presenceMonitoringEnabled), presenceMonitoringEnabled));
-
-        return entityManager.createQuery(cr).getResultStream();
+        return resourceRepository.findAllByPresenceMonitoringEnabled(presenceMonitoringEnabled).stream();
```
jjbuchan: I tried adding `findAllByPresenceMonitoringEnabledAsStream` to the repository but it didn't recognize `AsStream` even though I'm sure I saw that in some examples the other day.

jjbuchan: I'm also not certain we're even using this method right now, but it didn't hurt to leave it in.

itzg: I'm curious if you tried `...AndStream` rather than `...AsStream`? 

https://docs.spring.io/spring-data/jpa/docs/current/reference/html/#repositories.query-streaming

src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
```diff
@@ -209,8 +157,8 @@ public Resource getResource(String tenantId, String resourceId) {
      * @throws ResourceAlreadyExists
      */
     public Resource createResource(String tenantId, @Valid ResourceCreate newResource) throws IllegalArgumentException, ResourceAlreadyExists {
-        Resource existing = getResource(tenantId, newResource.getResourceId());
-        if (existing != null) {
+        Optional<Resource> existing = getResource(tenantId, newResource.getResourceId());
+        if (existing.isPresent()) {
```
itzg: It seems like repositories should support an "exists" style, but this section didn't clarify how the `Exists` keyword was used:

https://docs.spring.io/spring-data/jpa/docs/current/reference/html/#_supported_query_keywords

jjbuchan: That's a good list to see.  They have an example in the docs: `  boolean existsById(ID primaryKey);`

I'll try use that instead.

src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
jjbuchan: Neither works.  I assume it's not "special" since they only use it in their custom query.

PR #28, opened 2019-05-10
Updating unit test for use of underscore label delimiter

itzg:
# What

Due to https://github.com/racker/salus-telemetry-model/pull/46 the label namespace delimiter is now an underscore. The unit test was failing since a prefix of "system." was circumventing the validation that disallows "system_".


itzg:
Needs to be merged after https://github.com/racker/salus-telemetry-model/pull/46

PR #31, opened 2019-05-21
ReattachedEnvoyResourceEvent when Envoy attaches to existing resource

itzg:
# Resolves

https://jira.rax.io/browse/SALUS-360

# What

Extends the processing of Envoy attachment events by producing a new `ReattachedEnvoyResourceEvent` event introduced by https://github.com/racker/salus-telemetry-model/pull/52 when the resource already existed. That event will be produced in addition to the resource change event, if there was a change in the resource.

Design background is provided in https://github.com/racker/salus-telemetry-bundle/pull/50

## How to test

Unit tests updated and added.

### DIFF
--- a/src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
+++ b/src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
@@ -23,12 +23,19 @@ import com.rackspace.salus.resource_management.web.model.ResourceCreate;
 import com.rackspace.salus.resource_management.web.model.ResourceUpdate;
 import com.rackspace.salus.telemetry.errors.ResourceAlreadyExists;
 import com.rackspace.salus.telemetry.messaging.AttachEvent;
+import com.rackspace.salus.telemetry.messaging.ReattachedEnvoyResourceEvent;
 import com.rackspace.salus.telemetry.messaging.ResourceEvent;
 import com.rackspace.salus.telemetry.model.LabelNamespaces;
 import com.rackspace.salus.telemetry.model.NotFoundException;
 import com.rackspace.salus.telemetry.model.Resource;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Optional;
+import java.util.Set;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
@@ -268,6 +275,14 @@ public class ResourceManagement {
         if (existing.isPresent()) {
             log.debug("Found existing resource related to envoy: {}", existing.toString());
             updateEnvoyLabels(existing.get(), labels);
+
+            kafkaEgress.sendResourceEvent(
+                new ReattachedEnvoyResourceEvent()
+                    .setEnvoyId(attachEvent.getEnvoyId())
+                    .setTenantId(tenantId)
+                    .setResourceId(resourceId)
+            );
+
         } else {
             log.debug("No resource found for new envoy attach");
             Resource newResource = new Resource()
--- a/src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
+++ b/src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
@@ -376,19 +376,6 @@ public class ResourceManagement {
     }
 
     /**
-     * Publish a resource event to kafka for consumption by other services.
-     * @param resource The updated resource the operation was performed on.
-     */
-    private void publishResourceEvent(Resource resource) {
-        ResourceEvent event = new ResourceEvent();
-        event.setResourceId(resource.getResourceId());
-        event.setTenantId(resource.getTenantId());
-
-        log.debug("Publishing resource event: {}", event);
-        kafkaEgress.sendResourceEvent(event);
-    }
-
-    /**
      * This can be used to force a presence monitoring change if it is currently running but should not be.
      *
      * If the resource no longer exists, we will still send an event in case presence monitoring is still active.
--- a/src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
+++ b/src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
@@ -23,7 +23,6 @@ import com.rackspace.salus.resource_management.web.model.ResourceCreate;
 import com.rackspace.salus.resource_management.web.model.ResourceUpdate;
 import com.rackspace.salus.telemetry.errors.ResourceAlreadyExists;
 import com.rackspace.salus.telemetry.messaging.AttachEvent;
-import com.rackspace.salus.telemetry.messaging.ReattachedEnvoyResourceEvent;
 import com.rackspace.salus.telemetry.messaging.ResourceEvent;
 import com.rackspace.salus.telemetry.model.LabelNamespaces;
 import com.rackspace.salus.telemetry.model.NotFoundException;
@@ -36,7 +35,6 @@ import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Optional;
 import java.util.Set;
-import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import javax.validation.Valid;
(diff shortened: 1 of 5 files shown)


src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
```diff
@@ -268,6 +275,14 @@ public void handleEnvoyAttach(AttachEvent attachEvent) {
         if (existing.isPresent()) {
             log.debug("Found existing resource related to envoy: {}", existing.toString());
             updateEnvoyLabels(existing.get(), labels);
+
+            kafkaEgress.sendResourceEvent(
+                new ReattachedEnvoyResourceEvent()
+                    .setEnvoyId(attachEvent.getEnvoyId())
+                    .setTenantId(tenantId)
+                    .setResourceId(resourceId)
+            );
```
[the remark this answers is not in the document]
jjbuchan: OK, so it looks like a resource event is sent whenever a resource is created or labels are updated, which then triggers mon-mgmt to send monitors down the envoy.

I know I wrote the initial stuff for this, but I'm wondering if `updateEnvoyLabels` should be modified and all the events just get sent within this method.  The more we add the more things are getting confusing as they jump around.

I'm not yet following why we want to send two events to mon-mgmt when this one event is happening.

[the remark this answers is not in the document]
itzg: Yeah, we can definitely chat but had hoped the PR https://github.com/racker/salus-telemetry-bundle/pull/50 was a way to chat before writing all the code.

The original event remains a resource change event and does not have a need to convey envoy information. The new event has a distinct purpose that avoids bloating the other event/scenario. 

The monitor management queries for the affected resource and only acts upon changes. It currently has no way to differentiate between a resource that didn't change and the attached envoy of that resource.

I don't think the additional event is that complicated and in fact the "larger" amount of work was the reconciliation of bound monitors over in https://github.com/racker/salus-telemetry-monitor-management/pull/38 , which happens regardless of event structures.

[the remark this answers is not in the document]
itzg: Good point about

> The resource existing already does not mean the envoy has previously been connected.

What's a good way to detect the pre-existence of an envoy attached as a resource?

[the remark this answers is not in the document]
jjbuchan: I've been struggling to follow these changes while also trying to follow the ongoing meeting.

Ended up looking into things more to figure out exactly what is going on:

**handleResourceChangeEvent**
1. Gets the existing monitors bound to a resource prior to change
2. Get the monitors that should be bound to it after the label change
3. Determines monitors to unbind and does that by deleting those bound monitors from the db, decrementing counts, adds to list of changes
4. Determines if it needs to refresh existing monitors or add new ones, and then adds those to the list of changes
5. Sends changes to ambassador. One event for each unique envoy id.

**handleReattachedEnvoy**
1. Gets all local bound monitors for the resource
2. Gets the envoy ids these were tied to (should really only be 1 but accounts for many)
3. Updates the envoy id to the new one
4. Sends bound events for old ids and new id (one for each id)

-----

I wonder if ResourceMgmt should send a single event and based on the details of that MonMgmt will perform the relevant action.  This would eliminate the chance of us sending both update and attach events to MonitorMgmt, which is currently possible and could maybe cause race conditions?  This would ensure that we're not sending redundant MonitorBoundEvents from MonitorMgmt.

`updateEnvoyLabels` could return a boolean about whether it was updated or not.  I now remember the initial implementation involved sending more information in the `ResourceEvent`... since that has gone, I think it's easier to now send those events in `handleEnvoyAttach`.

> What's a good way to detect the pre-existence of an envoy attached as a resource?

We should probably start storing envoy connection history in sql?  I'm not sure if we should actually have a history of connections with timestamps and envoyIds or if we just need a `lastAgentConnectionTimestamp` on the Resource? Currently etcd can be queried to see the recently active envoy, but not anything historical.  I kinda like the idea of storing the last 10 connections for a resource - it could help with troubleshooting.

itzg:
@jjbuchan I'm going to think about it some more, but as I was making the changes to consolidate down to just `ResourceEvent` and introduce a resource-envoy attachment table I realized it was adding even more complexity and it potentially causes excessive queries on the Monitor Management side to process the label selectors. Even though I tend not to like pre-optimizing, Envoy reattachment without any other resource changes is common enough that it is worth optimizing that specific flow. We can chat more about my findings tomorrow.

What I did instead was at least the fix the situation you found where it can incorrectly think an Envoy reattached even when the resource was only created via the API.

itzg:
To give some more context to my previous comment, resource change events purposely provide no hints as to what changed about the resource, which is great to avoid race conditions of concurrent label changes to a resource. Where it doesn't work well IMHO, is in this case where only the Envoy is re-attaching to the resource and nothing else changed about the resource.

If this sent a `ResourceEvent` with every envoy re-attachment, `handleResourceEvent` of `MonitorManagement` has to run every time since it is blind about what actually changed about the resource:

https://sourcegraph.com/github.com/racker/salus-telemetry-monitor-management/-/blob/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java#L642

I haven't benchmarked it, but that method is probably one of our most intensive methods in the whole system. It looks up bound monitors, it does the label selector query, it calls back to the resource API, and more. If the resource itself didn't actually change, all of that would have executed just so that it could figure out just the envoyId field of the `BoundMonitor`s needed to change and send out monitor bound events.

Speaking of which, one of the concerns was the inter-leaving of an actual resource change event along with a re-attachment event. The worst that'll happen is that redundant monitor bound events will go out for a given envoy. The ambassador reconciles away those redundant updates via

https://sourcegraph.com/github.com/racker/salus-telemetry-ambassador/-/blob/src/main/java/com/rackspace/salus/telemetry/ambassador/services/EnvoyRegistry.java#L399:54-399:72

That itself, is an overhead, but a cheaper overhead than the `handleResourceEvent` method.

Imagine an entire DC's worth of Envoy have lost connectivity to our Ambassadors in GCP. When those Envoy's all re-attach there would have been a stampede of `handleResourceEvent` calls due to the resource change events; however, as implemented, it is limited to the purposely tuned `handleReattachedEnvoy` method

https://github.com/racker/salus-telemetry-monitor-management/pull/38/files#diff-9f3caf1ed599810e6ed74b7a5556cd14R693

src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
```diff
@@ -267,14 +274,30 @@ public void handleEnvoyAttach(AttachEvent attachEvent) {
 
         if (existing.isPresent()) {
             log.debug("Found existing resource related to envoy: {}", existing.toString());
-            updateEnvoyLabels(existing.get(), labels);
+
+            final Resource existingResource = updateEnvoyLabels(existing.get(), labels);
+
+            if (existingResource.isAssociatedWithEnvoy()) {
```
itzg: BTW, I have done end-to-end testing to verify this works.

jjbuchan:
> If this sent a ResourceEvent with every envoy re-attachment, handleResourceEvent of MonitorManagement has to run every time since it is blind about what actually changed about the resource:

I was thinking it would be modified to include fields about whether the labels were updated or not.  If they weren't it does the simple tasks, it they were then it does the label matching steps.

> resource change events purposely provide no hints as to what changed about the resource

In this case I think a small hint would be beneficial.  It's still not providing the details, it's just consolidating the messages into one earlier rather than having multiple services act on something that isn't needed.  Knowingly choosing to send messages you know will be irrelevant (and may add to confusion/complexity when troubleshooting) seems like a bad idea when it can be avoided.

itzg:
@jjbuchan ok, I think the addition of hints to the `ResourceEvent` would arrive at a good compromise. 

What's funny is that

> Knowingly choosing to send messages you know will be irrelevant (and may add to confusion/complexity when troubleshooting) seems like a bad idea

is exactly the reason I was adamant about the two distinct events :)

itzg:
@jjbuchan ready for re-review

src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
```diff
@@ -67,12 +72,21 @@ public ResourceManagement(ResourceRepository resourceRepository,
...
+     * @param labelsChanged
+     * @param reattachedEnvoyId
      * @return
      */
-    public Resource saveAndPublishResource(Resource resource) {
+    public Resource saveAndPublishResource(Resource resource, boolean labelsChanged,
+                                           String reattachedEnvoyId) {
         log.debug("Saving resource: {}", resource);
         resourceRepository.save(resource);
-        publishResourceEvent(resource);
```
jjbuchan: We can either delete `private void publishResourceEvent(Resource resource)` or move some of the logic back there.  Either works for me, but currently it's not being called anymore.

jjbuchan: well.... i probably do have a preference for it to exist, but am happy to leave it to you.  For some reason I just don't like seeing `kafkaEgress` requests scattered around.

itzg: I'll just remove `publishResourceEvent` at this point.

itzg: I changed my mind and introduced a new `publishResourceEvent` that takes the `ResourceEvent` rather than `Resource`...and that made me feel better.

PR #32, opened 2019-05-21
Tweak log message

jjbuchan:
Small change I made while working on other things.

This is an `Optional` field, so get it first before logging.

### DIFF
--- a/src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
+++ b/src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
@@ -266,7 +266,7 @@ public class ResourceManagement {
         Optional<Resource> existing = getResource(tenantId, resourceId);
 
         if (existing.isPresent()) {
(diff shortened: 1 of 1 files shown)


src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
```diff
@@ -266,7 +266,7 @@ public void handleEnvoyAttach(AttachEvent attachEvent) {
         Optional<Resource> existing = getResource(tenantId, resourceId);
 
         if (existing.isPresent()) {
-            log.debug("Found existing resource related to envoy: {}", existing.toString());
+            log.debug("Found existing resource related to envoy: {}", existing.get().toString());
```
itzg: or just

```suggestion
            log.debug("Found existing resource related to envoy: {}", existing.get());
```

since slf4j will invoke `toString` when the debug log level is active.

PR #33, opened 2019-06-03
Add ResourceDTO and add creation/update timestamps

jjbuchan:
# What

1. Moves `Resource.java` from `salus-models` and adds it here.
1. Introduces a `ResourceDTO` to separate what is being returned to users from the entity stored in mysql.
1. Adds creation and update timestamps to the model repo.

# How

1. Copy/Paste
1. Basically copied the same method we use elsewhere.  Added a `toDTO` method.  PRs incoming for changing this in places that use the ResourceAPI.
1. `@CreationTimestamp` & `@UpdateTimestamp` do all the work.

## How to test

Run tests or try manually updating resources via the api.

# Why

1. To make things consistent with other services and to exclude the Resource object from other services.  They should not interact with the db entity.
1. Having a DTO allows us to return different fields to the customer than we might have specified in the db.  i.e. admin users can see more than customers.
1. Timestamps on objects is helpful for debugging.

ResourceApiTest.java  and ResourceApiControllerTest.java were basically trying to do the same thing I consolidated them.  Reviewing that might be a pain.

# TODO

Remove `Resource` and `ResourceAlreadyExists` from the model repo.
Need to figure out what to do with ValidLabelKeysValidatorTest.java.  It relies on `Resource` and can no longer access it.


src/test/java/com/rackspace/salus/resource_management/TestUtils.java
```diff
@@ -0,0 +1,33 @@
...
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import org.springframework.core.io.ClassPathResource;
+import org.springframework.util.FileCopyUtils;
+
+public class TestUtils {
+
+  public static String readContent(String resource) throws IOException {
```
GeorgeJahad: should this be in libs/common?

src/test/java/com/rackspace/salus/resource_management/web/controller/ResourceApiControllerTest.java
```diff
@@ -14,75 +14,328 @@
...
+    Stream<Resource> resourceStream = resources.stream();
+
+    when(resourceManagement.getResources(true))
+        .thenReturn(resourceStream);
+
+    mockMvc.perform(get(url))
+        .andDo(print())
+        .andExpect(status().isOk())
+        .andExpect(content().contentTypeCompatibleWith("text/event-stream;charset=UTF-8"))
+        .andExpect(content().string(stringContainsInOrder(expectedData)));
```
GeorgeJahad: why do you check the content.string() in this test but not the previous ones?

itzg:
Regarding `ValidLabelKeysValidatorTest`, I'd say just switch over to fully using that inner `DataWithLabels` data class. I originally used `Resource` in the test to be more "realistic" with the unit test.

src/main/java/com/rackspace/salus/resource_management/entities/lombok.config
```diff
@@ -0,0 +1,2 @@
+config.stopBubbling = true
```
itzg: In monitor-management I just ended up moving this to the "application" package level, which would be `com.rackspace.salus.resource_management` in this case.

src/main/java/com/rackspace/salus/resource_management/errors/ResourceAlreadyExists.java
```diff
@@ -0,0 +1,21 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.rackspace.salus.resource_management.errors;
+
+public class ResourceAlreadyExists extends RuntimeException {
```
itzg: I wonder if we should just be using `AlreadyExistsException`.

src/main/java/com/rackspace/salus/resource_management/web/client/ResourceApiClient.java
```diff
@@ -17,7 +17,8 @@
 package com.rackspace.salus.resource_management.web.client;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
-import com.rackspace.salus.telemetry.model.Resource;
+import com.rackspace.salus.resource_management.entities.Resource;
```
itzg: Since the client jar doesn't include `.entities.Resource` can you check and remove this import?

src/main/java/com/rackspace/salus/resource_management/web/controller/ResourceApiController.java
```diff
@@ -73,22 +77,24 @@ public ResourceApiController(ResourceManagement resourceManagement, TaskExecutor
     }
 
     @GetMapping("/resources")
+    @JsonView(View.Admin.class)
     @ApiOperation(value = "Gets all Resources irrespective of Tenant")
-    public Page<Resource> getAll(@RequestParam(defaultValue = "100") int size,
+    public PagedContent<ResourceDTO> getAll(@RequestParam(defaultValue = "100") int size,
```
itzg: Thank you for switching to `PagedContent` -- the json rendering of `Page` is so noisy.

src/test/java/com/rackspace/salus/resource_management/TestUtils.java
itzg: We definitely should find a better home for this test method that all of these modules will end up declaring. The downside to putting in libs/common is that technically we wouldn't want `TestUtils` to remain in the final, production jars -- we'd want it to be pulled in via a test scoped dependency. So...the "proper" solution is a new `libs/salus-test` module like how Spring has [spring-test](https://mvnrepository.com/artifact/org.springframework/spring-test).

I wrote up https://jira.rax.io/browse/SALUS-383 for me to work on this.

src/test/java/com/rackspace/salus/resource_management/web/controller/ResourceApiControllerTest.java
```diff
@@ -14,75 +14,328 @@
...
+  }
+
+  @Test
+  public void testNoResourceFound() throws Exception {
+    when(resourceManagement.getResource(anyString(), anyString()))
+        .thenReturn(Optional.empty());
+
+    String tenantId = RandomStringUtils.randomAlphabetic( 8 );
+    String resourceId = RandomStringUtils.randomAlphabetic( 8 );
+    String url = String.format("/api/tenant/%s/resources/%s", tenantId, resourceId);
```
itzg: Instead of this, you could use the version of [get](https://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/test/web/servlet/request/MockMvcRequestBuilders.html#get-java.lang.String-java.lang.Object...-) that accepts `uriVars` like the RestTemplate calls.

src/test/java/com/rackspace/salus/resource_management/web/controller/ResourceApiControllerTest.java
```diff
@@ -14,75 +14,328 @@
...
         "/api/tenant/{tenantId}/resources/{resourceId}",
         "t-1", "r-1"
     ).accept(MediaType.APPLICATION_JSON))
         .andExpect(status().isOk())
-        .andExpect(content().json(objectMapper.writeValueAsString(expectedResource)));
+        .andExpect(content().json(
+            // id field should not be returned
+            readContent("ResourceApiControllerTest/single_public_resource.json"), true));
+
+  }
```
itzg: For completeness, there should be Mockito verify's of each mocked method, in this case `resourceManagement.getResource`. That way it can confirm the controller did the expected backend thing.

src/main/java/com/rackspace/salus/resource_management/errors/ResourceAlreadyExists.java
jjbuchan: We currently have numerous `AlreadyExist` errors.  I'm good with switching them all to `AlreadyExistsException`.  I'll do that in another PR.

src/test/java/com/rackspace/salus/resource_management/web/controller/ResourceApiControllerTest.java
jjbuchan: Two things
1. This uses `.string` rather than `.json`.  It does that because it's the stream rather than a single response so it was easier to do it that way.
2. I removed some of the `content().json` checks from other tests because adding views to endpoints leads to the old method of testing not being possible.  We'd have to pre-define what we expect to see and have all the variations stored in json files.  We should also add ResourceDTO tests in addition to this.  I did all this for the ZoneAPI and ZoneDTO, but held off here since the JsonViews stuff isn't really solidified yet.
We do still have some tests for smaller requests that check the body returned, so I figured that'd get us by for now.

PR #43, opened 2019-07-16
Add tenant APIs to query labels and metadata keys in use

itzg:
# Resolves

https://jira.rax.io/browse/SALUS-450

# What

Adds REST API to query for the labels (key-value) and metadata (keys-only) currently in use on their resources. This will enable UIs to auto-suggest labels and metadatas for resource create and update operations.

## How to test

Unit tests expanded to cover new operations

# Requires

https://github.com/racker/salus-telemetry-model/pull/58


src/main/java/com/rackspace/salus/resource_management/web/controller/ResourceApiController.java
```diff
@@ -157,4 +161,21 @@ public void delete(@PathVariable String tenantId,
...
+    return resourceManagement.getTenantResourceLabels(tenantId);
+  }
+
+  @GetMapping("/tenant/{tenantId}/resource-metadata-keys")
+  @JsonView(View.Public.class)
+  public List<String> getResourceMetadataKeys(@PathVariable String tenantId) {
+    return resourceManagement.getTenantResourceMetadataKeys(tenantId);
+  }
+
+  @GetMapping("/tenant/{tenantId}/resource-label-namespaces")
```
itzg: @jjbuchan your question about account policy overlaps made me realize that this particular API call might need to belong over there. It's not really specific to resources, but resource management seemed the best place for this query for now.

src/test/resources/ResourceApiControllerTest/label_namespaces.json
```diff
@@ -0,0 +1 @@
+["agent", "system"]
```
itzg: @FarajiA this and the next two files are examples of the response structure for each of the new API operations

src/main/java/com/rackspace/salus/resource_management/entities/Resource.java
```diff
@@ -52,6 +54,12 @@
         // also enables indexed query by tenant ID
         @UniqueConstraint(columnNames = {"tenant_id", "resource_id"})
     })
+@NamedQueries({
+    @NamedQuery(name = "resourceGetDistinctLabels",
+    query = "select distinct entry(r.labels) from Resource r where r.tenantId = :tenantId"),
+    @NamedQuery(name = "resourceGetMetadata",
+    query = "select r.metadata from Resource r where r.tenantId = :tenantId")
```
jjbuchan: Not to get into another big discussion, but would these not work as normal repository queries since it is getting a value that is directly on the main Resource object?

It seems like we decided the previous one kept warning because it was an object within an object within an object etc.

src/test/resources/ResourceApiControllerTest/resource_labels.json
```diff
@@ -0,0 +1,14 @@
+{
+  "key1": [
```
FarajiA: Okay so querying the endpoint for a tenant will give all their distinct resource labels. Does "key1" represent a resource identifier or a label key, if label key when will this ever have more than one value?

src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
```diff
@@ -439,6 +449,40 @@ private void removePresenceMonitoring(String tenantId, String resourceId) {
...
+    /*
+    NOTE: since metadata is stored as a json text column there's not really an efficient way
+    to pull out just the metadata keys. As a result, the following iterates over just the
+    metadata of the tenant's resources and picks out the distinct keys.
+     */
+    @SuppressWarnings("unchecked")
+    final Stream<Map<String,String>> resultStream = entityManager
+        .createNamedQuery("resourceGetMetadata")
+        .setParameter("tenantId", tenantId)
+        .setHint("org.hibernate.fetchSize", resourceManagementProperties.getResourceMetadataFetchSize())
```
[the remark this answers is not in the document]
itzg: I'm assuming it gets fed into this aspect of JDBC cursor queries https://docs.oracle.com/cd/E18283_01/java.112/e16548/resltset.htm#i1023619

src/main/java/com/rackspace/salus/resource_management/entities/Resource.java
itzg: These ones probably work technically as repository queries, but I'm going to double down on the philosophical stance and the IntelliJ warning that: repository queries should return the entity, collections of the entity, or projections of the entity. The last rule can be abused to accommodate `resourceGetMetadata`, but the spirit of this particular query disregards the individual resource entities and just extracts the "aggregate" across them.

What I am going to do though is use a better naming convention on these, such as `Resource.getMetadata`.

src/test/resources/ResourceApiControllerTest/resource_labels.json
itzg: Oh yeah, these fake values are confusing. I'm going to change this to be

```json
{
  "agent_discovered_os": [
    "linux",
    "darwin",
    "windows"
  ],
  "agent_discovered_arch": [
    "amd64",
    "386"
  ],
  "cluster": [
    "dev",
    "prod"
  ]
}
```

where it's easier to see that the resource identifiers are not included in the response but rather those are the label keys. It's almost more apparent why there would be more than one value per key.

FarajiA: Gotcha

PR #47, opened 2019-07-30
Add endpoint to get all tenants with at least one resource

jjbuchan:
# What

Returns all distinct tenant ids that the resource management service is aware of.

# How

Simple db query + adds api endpoints + tests.

## How to test

Run tests!

# Why

Policy Management will query this when a policy change occurs and then send out a single `PolicyEvent` for each tenant.  This is so the monitor management service (and other in future) will be able to consume these one at a time, deciding what action needs to happen for an individual tenant, rather than being overloaded by trying to handle all tenants at once.

### DIFF
--- a/src/main/java/com/rackspace/salus/resource_management/entities/Resource.java
+++ b/src/main/java/com/rackspace/salus/resource_management/entities/Resource.java
@@ -58,7 +58,9 @@ import org.hibernate.validator.constraints.NotBlank;
     @NamedQuery(name = "Resource.getDistinctLabels",
     query = "select distinct entry(r.labels) from Resource r where r.tenantId = :tenantId"),
     @NamedQuery(name = "Resource.getMetadata",
-    query = "select r.metadata from Resource r where r.tenantId = :tenantId")
+    query = "select r.metadata from Resource r where r.tenantId = :tenantId"),
+    @NamedQuery(name = "Resource.getAllDistinctTenants",
+    query = "select distinct r.tenantId from Resource r")
 })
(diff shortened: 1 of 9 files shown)


src/main/java/com/rackspace/salus/resource_management/entities/Resource.java
```diff
@@ -58,7 +58,9 @@
     @NamedQuery(name = "Resource.getDistinctLabels",
     query = "select distinct entry(r.labels) from Resource r where r.tenantId = :tenantId"),
     @NamedQuery(name = "Resource.getMetadata",
-    query = "select r.metadata from Resource r where r.tenantId = :tenantId")
+    query = "select r.metadata from Resource r where r.tenantId = :tenantId"),
+    @NamedQuery(name = "Resource.getAllDistinctTenants",
```
itzg: I had already forgotten I had just added named queries to `Resource`, so at least yours wasn't a one-off named query.

src/main/java/com/rackspace/salus/resource_management/repositories/ResourceRepository.java
```diff
@@ -21,6 +21,7 @@
 import java.util.Optional;
 import org.springframework.data.domain.Page;
 import org.springframework.data.domain.Pageable;
+import org.springframework.data.jpa.repository.Query;
```
itzg: import isn't needed, but can be fixed next time someone is in this file.

jjbuchan: Just you wait.  I'll get one in there somehow!

PR #49, opened 2019-08-09
Some changes that can happen due to using a shared database

jjbuchan:
# What

Some cleanup since not all these things are needed anymore.

Had to comment out the JsonViews section for tenantId - we probably need to do this for numerous other places too.  What we really need is role based views.  This topic could be for eng sync unless I'm able to look and find a good alternative option.

Without digging to much into it this might be helpful - https://www.baeldung.com/spring-security-role-filter-json


itzg:
> Without digging to much into it this might be helpful - https://www.baeldung.com/spring-security-role-filter-json

Yeah, I like the look of that solution.

PR #54, opened 2019-09-13
handling jdbc exceptions and kafka exceptions through the RestExcepti…

harageth:
# Resolves

https://jira.rax.io/browse/SALUS-433

# What

This PR is trying to catch the error when the API nodes are up but a service behind the API node isn't. 

# How

I am catching the particular error that gets thrown when another service isn't running. 

## How to test

Start this service after starting all of the docker containers. 
After this has started shut down mysql.
Send a creation request in.
Bring mysql back up and shut down kafka.
send the same creation request.

# Why

For unknown reasons Spring Boot doesn't seem to appreciate using the configuration options for not including the exception or the stack trace.

Instead I decided to utilize the pattern we are already using for catching exceptions and responding with HTTP status codes to give a friendly message.

### DIFF
--- a/src/main/java/com/rackspace/salus/resource_management/services/KafkaEgress.java
+++ b/src/main/java/com/rackspace/salus/resource_management/services/KafkaEgress.java
@@ -16,10 +16,12 @@
 
 package com.rackspace.salus.resource_management.services;
 
+import com.rackspace.salus.common.errors.RuntimeKafkaException;
 import com.rackspace.salus.common.messaging.KafkaMessageKeyBuilder;
 import com.rackspace.salus.common.messaging.KafkaTopicProperties;
 import com.rackspace.salus.telemetry.messaging.KafkaMessageType;
 import com.rackspace.salus.telemetry.messaging.ResourceEvent;
+import java.util.concurrent.ExecutionException;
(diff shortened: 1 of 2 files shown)


src/main/java/com/rackspace/salus/resource_management/services/KafkaEgress.java
```diff
@@ -46,6 +48,13 @@ public void sendResourceEvent(ResourceEvent event) {
 
         log.debug("Sending event={} on topic={}", event, topic);
         final String key = KafkaMessageKeyBuilder.buildMessageKey(event);
-        kafkaTemplate.send(topic, key, event);
+
+        try {
+            kafkaTemplate.send(topic, key, event).get();
+        } catch (InterruptedException e) {
```
itzg: FYI, starting with Java 7 you can [combine caught exceptions](https://docs.oracle.com/javase/7/docs/technotes/guides/language/catch-multiple.html) using:

```
catch (InterruptedException|ExecutionException e)
```

You'll probably find IntelliJ is suggesting that fix.

harageth: Weirdly it didn't and I even had Intellij generate the try/catch for me

PR #55, opened 2019-09-18
fixed the kafka tests

harageth:
Fixed the tests because I didn't realize it was failing

### DIFF
--- a/src/test/java/com/rackspace/salus/resource_management/services/KafkaEgressTest.java
+++ b/src/test/java/com/rackspace/salus/resource_management/services/KafkaEgressTest.java
@@ -16,7 +16,10 @@
 
 package com.rackspace.salus.resource_management.services;
 
+import static org.mockito.ArgumentMatchers.any;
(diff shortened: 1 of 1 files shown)


src/test/java/com/rackspace/salus/resource_management/services/KafkaEgressTest.java
```diff
@@ -43,6 +48,10 @@ public void setUp() {
 
   @Test
   public void testSendResourceEvent() {
+    SettableListenableFuture<SendResult<String, Object>> future = new SettableListenableFuture();
+    future.set(null);
+    when(kafkaTemplate.send(anyString(), anyString(), any())).thenReturn(future);
```
jjbuchan: FYI, you can probably use `thenReturn(CompletableFuture.completedFuture(null))`.  We do that in a few places.

harageth: Ohhhh gosh I didn't even see that static function. I tried some similar things with no success.

PR #56, opened 2019-09-26
Logical OR processing for resources

harageth:
# Resolves

https://jira.rax.io/browse/SALUS-606

# What

Support logical OR as well as logical AND processing for matching labels. The label matching in here is from monitors to resources so we dont need to combine the results together since the monitor will tell us what sort of processing we need to do. 

# How

I duplicated the AND query and modified it for handling OR processing. 

## How to test

Run the unit tests. The functions that are testing this end in "UsingOr"

# Why

We considered handling this in a hybrid approach through JPA or Hibernate. Neither of those places seemed to have a good way of querying map values. 

There was also a consideration for doing this through a hybrid approach but that would still require the writing of the OR based query and then handling the rest of the processing server side. Since we already have the AND query it seems more logical to me to just handle this all at query time.

### DIFF
--- a/src/main/java/com/rackspace/salus/resource_management/web/client/ResourceApiClient.java
+++ b/src/main/java/com/rackspace/salus/resource_management/web/client/ResourceApiClient.java
@@ -18,6 +18,7 @@ package com.rackspace.salus.resource_management.web.client;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.rackspace.salus.resource_management.web.model.ResourceDTO;
+import com.rackspace.salus.telemetry.model.LabelSelectorMethod;
 import com.rackspace.salus.telemetry.model.PagedContent;
 import java.io.BufferedReader;
 import java.io.IOException;
@@ -98,13 +99,16 @@ public class ResourceApiClient implements ResourceApi {
   }
 
   @Override
-  public List<ResourceDTO> getResourcesWithLabels(String tenantId, Map<String, String> labels) {
-    String endpoint = "/api/tenant/{tenantId}/resources-by-label/AND";
+  public List<ResourceDTO> getResourcesWithLabels(String tenantId, Map<String, String> labels, LabelSelectorMethod labelSelector) {
+    String endpoint = "/api/tenant/{tenantId}/resources-by-label/{logicalOperator}";
     UriComponentsBuilder uriComponentsBuilder = UriComponentsBuilder.fromUriString(endpoint);
     for (Map.Entry<String, String> e : labels.entrySet()) {
       uriComponentsBuilder.queryParam(e.getKey(), e.getValue());
     }
-    String uriString = uriComponentsBuilder.buildAndExpand(tenantId).toUriString();
+    //String uriString = uriComponentsBuilder.buildAndExpand(tenantId).toUriString();
+    Object[] items = {tenantId, labelSelector};
+
+    String uriString = uriComponentsBuilder.buildAndExpand(items).toUriString();
     ResponseEntity<PagedContent<ResourceDTO>> resp = restTemplate.exchange(
         uriString,
         HttpMethod.GET,
--- a/src/main/java/com/rackspace/salus/resource_management/web/client/ResourceApiClient.java
+++ b/src/main/java/com/rackspace/salus/resource_management/web/client/ResourceApiClient.java
@@ -105,10 +105,8 @@ public class ResourceApiClient implements ResourceApi {
     for (Map.Entry<String, String> e : labels.entrySet()) {
       uriComponentsBuilder.queryParam(e.getKey(), e.getValue());
     }
-    //String uriString = uriComponentsBuilder.buildAndExpand(tenantId).toUriString();
-    Object[] items = {tenantId, labelSelector};
 
-    String uriString = uriComponentsBuilder.buildAndExpand(items).toUriString();
+    String uriString = uriComponentsBuilder.buildAndExpand(tenantId, labelSelector).toUriString();
     ResponseEntity<PagedContent<ResourceDTO>> resp = restTemplate.exchange(
         uriString,
         HttpMethod.GET,
--- a/src/main/java/com/rackspace/salus/resource_management/web/client/ResourceApiClient.java
+++ b/src/main/java/com/rackspace/salus/resource_management/web/client/ResourceApiClient.java
@@ -99,7 +99,7 @@ public class ResourceApiClient implements ResourceApi {
 
   @Override
   public List<ResourceDTO> getResourcesWithLabels(String tenantId, Map<String, String> labels) {
-    String endpoint = "/api/tenant/{tenantId}/resources-by-label";
+    String endpoint = "/api/tenant/{tenantId}/resources-by-label/AND";
     UriComponentsBuilder uriComponentsBuilder = UriComponentsBuilder.fromUriString(endpoint);
     for (Map.Entry<String, String> e : labels.entrySet()) {
       uriComponentsBuilder.queryParam(e.getKey(), e.getValue());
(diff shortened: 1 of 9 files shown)


src/main/java/com/rackspace/salus/resource_management/web/controller/ResourceApiController.java
```diff
@@ -151,11 +152,11 @@ public void delete(@PathVariable String tenantId,
     resourceManagement.removeResource(tenantId, resourceId);
   }
 
-  @GetMapping("/tenant/{tenantId}/resources-by-label")
+  @GetMapping("/tenant/{tenantId}/resources-by-label/{logicalOperation}")
   @JsonView(View.Public.class)
   public PagedContent<ResourceDTO> getResourcesWithLabels(@PathVariable String tenantId,
-      @RequestParam Map<String, String> labels, Pageable pageable) {
-    return PagedContent.fromPage(resourceManagement.getResourcesFromLabels(labels, tenantId, pageable)
+      @RequestParam Map<String, String> labels, @PathVariable String logicalOperation, Pageable pageable) {
```
itzg: It feels like the logical operation should be a query parameter it is more like a modifier of the query rather than an identifier. If you agree, the good news is that it's as easy as switching to `@RequestParam` and removing the reference from the mapping URL.

src/main/resources/sql-queries/resource_label_matching_query.sql
```diff
@@ -24,4 +24,5 @@ WHERE
             HAVING COUNT(*) = :i
          )
    )
+GROUP BY resources.id
```
itzg: Just curious why the `GROUP BY` here in these two queries.

src/test/java/com/rackspace/salus/resource_management/ResourceManagementTest.java
```diff
@@ -745,6 +746,7 @@ public void testGetResourcesFromLabels_multipleMatches() {
         Page<Resource> resources = resourceManagement.getResourcesFromLabels(
             Collections.singletonMap("os", "linux"),
             "testGetResourcesFromLabels_multipleMatches",
+            LabelSelectorMethod.AND,
```
itzg: It doesn't look like there's any OR tests. I'd say pick one or two of these existing scenarios and add a similar one with an OR.

src/main/java/com/rackspace/salus/resource_management/web/controller/ResourceApiController.java
harageth: Yes that was actually something I meant to mention in the TODO section. I left it there because it felt easier for testing.

src/test/java/com/rackspace/salus/resource_management/ResourceManagementTest.java
harageth: That was something I still needed to do. I left that comment in the TODO section. 

Just wanted to make sure I wasn't so far out in left field with what I had done before I had done that.

src/main/java/com/rackspace/salus/resource_management/web/controller/ResourceApiController.java
itzg: You'll be glad to find it's pretty much the same for testing query params vs path variables.

src/test/java/com/rackspace/salus/resource_management/ResourceManagementTest.java
itzg: I'm sorry, I meant to delete this comment after re-reading the PR description.

src/main/resources/sql-queries/resource_label_matching_query.sql
harageth: Paranoia. I think I saw one result that had 12000 rows in it. Which would have been the join for all of the labels for the 3000 resources I had created. 

I could have just as easily achieved the same result by doing a DISTINCT in the SELECT. I will double check that this isn't needed.

src/main/java/com/rackspace/salus/resource_management/web/client/ResourceApiClient.java
```diff
@@ -98,13 +99,16 @@ public ResourceDTO getByResourceId(String tenantId, String resourceId) {
...
+    String endpoint = "/api/tenant/{tenantId}/resources-by-label/{logicalOperator}";
     UriComponentsBuilder uriComponentsBuilder = UriComponentsBuilder.fromUriString(endpoint);
     for (Map.Entry<String, String> e : labels.entrySet()) {
       uriComponentsBuilder.queryParam(e.getKey(), e.getValue());
     }
-    String uriString = uriComponentsBuilder.buildAndExpand(tenantId).toUriString();
+    //String uriString = uriComponentsBuilder.buildAndExpand(tenantId).toUriString();
+    Object[] items = {tenantId, labelSelector};
+
+    String uriString = uriComponentsBuilder.buildAndExpand(items).toUriString();
```
itzg: It's a vararg `Object...` so you don't need to create an array, just do:

```
    String uriString = uriComponentsBuilder.buildAndExpand(tenantId, labelSelector).toUriString();
```

src/main/java/com/rackspace/salus/resource_management/web/controller/ResourceApiController.java
```diff
@@ -152,11 +152,11 @@ public void delete(@PathVariable String tenantId,
     resourceManagement.removeResource(tenantId, resourceId);
   }
 
-  @GetMapping("/tenant/{tenantId}/resources-by-label/{logicalOperation}")
+  @GetMapping("/tenant/{tenantId}/resources-by-label/{logicalOperator}")
   @JsonView(View.Public.class)
   public PagedContent<ResourceDTO> getResourcesWithLabels(@PathVariable String tenantId,
-      @RequestParam Map<String, String> labels, @PathVariable String logicalOperation, Pageable pageable) {
-    return PagedContent.fromPage(resourceManagement.getResourcesFromLabels(labels, tenantId, LabelSelectorMethod.valueOf(logicalOperation), pageable)
+      @RequestParam Map<String, String> labels, @PathVariable String logicalOperator, Pageable pageable) {
```
itzg: If you declare the arg as type `LabelSelectorMethod` I think Spring MVC will take care of the enum conversion (and validation).

src/test/java/com/rackspace/salus/resource_management/ResourceManagementTest.java
```diff
@@ -665,6 +687,29 @@ public void testMatchResourceWithSupersetOfLabels() {
         assertEquals(resourceLabels, resources.getContent().get(0).getLabels());
     }
 
+    @Test
+    public void testMatchResourceWithSupersetOfLabelsDoingOr() {
+        final Map<String, String> resourceLabels = new HashMap<>();
+        resourceLabels.put("os", "DARWIN");
+        resourceLabels.put("env", "test");
```
itzg: You should make one of the two labels not have a matching value to confirm the "or" is actually working differently than an "and". For example, changing this to `"env", "prod"` should retain the same unit test expectations.

src/main/resources/sql-queries/resource_label_matching_OR_query.sql
```diff
@@ -0,0 +1,28 @@
...
+FROM
+   resources JOIN resource_labels AS rl
+WHERE
+   resources.id = rl.id
+   AND resources.id IN
+   (
+      SELECT inner_rl.id
+      FROM resource_labels AS inner_rl
+      WHERE
+         id IN
```
itzg: Probably should qualify this `id` mostly because I'm wondering if it uses `resources.id as id` from L2

harageth: Nice catch. It blended in with the IN on my text highlighting.

GeorgeJahad:
This looks pretty good, but I don't see a test case that shows that a bad match correctly returns nothing when usingOr.  Did I miss it?

itzg:
After this one gets merged to master then the monitor-management build on master needs to be manually re-run to resolve compilation error like

https://console.cloud.google.com/cloud-build/builds/cc949d2b-52ba-4dad-b26b-87cef9afc859?project=salus-220516&authuser=1&organizationId=767495606862

PR #58, opened 2019-10-08
Support unpaged, inter-service resources query

itzg:
# Resolves

https://jira.rax.io/browse/SALUS-642

# What

When creating a monitor for our ele resources in the dev cluster I noticed that only 20 bound monitors were getting created. It turns out that ever since combining the public and admin queries into one call, this operation has been "broken" due to a default page size.

# How

- Filter the pageable parameters out of the public API's `labels` request parameter catch-all map
- Add back an admin API call, but reuse the same service component's query method
- Fix the warning IntelliJ reports about multiple `TaskExecutor` beans

# How to test

Updated unit tests


jjbuchan:
I'm failing to see the commit but is this not the one that used to just request `Pageable.unpaged()`?

Can we just go back to something simpler like that instead of having to do all this stuff?

```
    final Map<String, String> resourceLabels = new HashMap<>(labels);
    resourceLabels.remove(springDataWebProperties.getPageable().getSizeParameter());
    resourceLabels.remove(springDataWebProperties.getPageable().getPageParameter());
```

itzg:
> I'm failing to see the commit but is this not the one that used to just request `Pageable.unpaged()`?
> 
> Can we just go back to something simpler like that instead of having to do all this stuff?

I didn't look at all for the previous version since it also just returned a list of resource IDs, which no longer is useful by itself since we need to render the monitor config, etc.

Without the `page` and `size` filtering the existing REST call gets those parameters leaked with a failure like:

```
Actual invocation has different arguments:
com.rackspace.salus.resource_management.services.ResourceManagement#0 bean.getResourcesFromLabels(
    {"size" = "101", "env" = "prod"},
    "t-1",
    AND,
    Page request [number: 0, size 101, sort: UNSORTED]
);
```

On the flip side, I tried using the one REST API call and passing `Integer.MAX_VALUE` for the `size` parameter, but the `Pageable` argument parsing imposes a limit (rightly so) on the size and maxes out at 2000, by default.

So my theory was, keep the public facing API call with the proper paging limits, but avoid leaking paging parameters into the labels. Add a new admin API call that does the simpler thing.

jjbuchan:
Looked over things again in "split view" and that made it clearer.

> but the Pageable argument parsing imposes a limit (rightly so) on the size and maxes out at 2000, by default.

Didn't know this.

itzg:
> > but the Pageable argument parsing imposes a limit (rightly so) on the size and maxes out at 2000, by default.
> 
> Didn't know this.

I didn't either until my unit test assertions kept failing with this mystery value of 2000 showing up in place of the max integer value I was giving :)

src/main/java/com/rackspace/salus/resource_management/web/controller/ResourceApiController.java
```diff
@@ -152,12 +162,38 @@ public void delete(@PathVariable String tenantId,
...
+                                                          Pageable pageable) {
+
+    // labels is a catch-all for request/query parameters, so need to strip out the pageable parameters
+    final Map<String, String> resourceLabels = new HashMap<>(labels);
+    resourceLabels.remove(springDataWebProperties.getPageable().getSizeParameter());
+    resourceLabels.remove(springDataWebProperties.getPageable().getPageParameter());
+
+    // and convert to unpaged when "one big page is requested" since the query logic handles this more directly
+    if (pageable.getPageNumber() == 0 && pageable.getPageSize() == Integer.MAX_VALUE) {
+      pageable = Pageable.unpaged();
```
jjbuchan: I still don't think I'm quite following this.

Why would we want customers to get a fully unpaged result?  Surely it's only our admin api that should do that?  The limit of 2000 seems more than enough for customers.

> the query logic handles this more directly

hmm?

itzg: Oops, I meant to remove that failed attempt to do the "unpaged" conversion.

jjbuchan: A test that shows customers cant get more than 2k might be worthwhile?

itzg: It's a Spring Data Commons enforced thing, but should be easy to add a unit test for that. Technically this limit is being imposed anywhere we're using `Pageable` request arguments.

jjbuchan: I was thinking more since this change would have overridden it :) But I don't mind too much due to the points you stated.

itzg: FYI it seems like [this section of docs](https://docs.spring.io/spring-data/data-commons/docs/current/reference/html/#core.web.basic.paging-and-sorting) should have described the max page size limit, but here's the source code

https://github.com/spring-projects/spring-data-commons/blob/2.1.x/src/main/java/org/springframework/data/web/PageableHandlerMethodArgumentResolver.java#L55

PR #61, opened 2019-11-12
Add create endpoint to api client

jjbuchan:
# What

Adds a create endpoint to the api client.

# How

I included `MultiValueMap<String, String> headers` so that this can be used against the public api and provide an `x-auth-token`.  I spoke to @itzg briefly about that before and if the need arises we might want to add that to all the api client methods (or at least any we add from here on) so that it is more of a comprehensive client that can be used for multiple purposes and not just internal service-to-service requests.

# Why

I'm utilizing this client in the migration app.

### DIFF
--- a/src/main/java/com/rackspace/salus/resource_management/web/client/ResourceApiClient.java
+++ b/src/main/java/com/rackspace/salus/resource_management/web/client/ResourceApiClient.java
@@ -16,7 +16,10 @@
 
 package com.rackspace.salus.resource_management.web.client;
 
+import static com.rackspace.salus.common.web.RemoteOperations.mapRestClientExceptions;
+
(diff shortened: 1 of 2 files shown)


src/main/java/com/rackspace/salus/resource_management/web/client/ResourceApiClient.java
```diff
@@ -64,6 +71,8 @@ public ResourceApi resourceApi(RestTemplateBuilder restTemplateBuilder) {
 @Slf4j
 public class ResourceApiClient implements ResourceApi {
 
+  private static final String SERVICE_NAME = "resource-management";
```
jjbuchan: This was copied from how monitor-management does it.

PR #65, opened 2020-01-22
Fixed resource management label query and added extra tests

harageth:
# Resolves

https://jira.rax.io/browse/SALUS-737

# What

Make sure that a resource with no labels will still match everything

# How

Alters the query to make sure that we are using an OUTER JOIN to find the resources that dont have any labels attached to them and still makes sure that we are only grabbing the resources for that tenant.

## How to test

Run the ResourceMangementTest's

# Why

It needs to be done at query time. We can't do it on the server because by the time we get there we would have already filtered out the resources that don't have labels.

# TODO

Everything should be done

### DIFF
--- a/src/test/java/com/rackspace/salus/resource_management/ResourceManagementTest.java
+++ b/src/test/java/com/rackspace/salus/resource_management/ResourceManagementTest.java
@@ -698,6 +698,7 @@ public class ResourceManagementTest {
         assertThat(resourceIds, containsInAnyOrder(create.getResourceId(), create2.getResourceId()));
     }
 
+    //This test is supposed to make sure that when matching resources we are still only returning the tenants requested
     public void testMatchResourceWithNoLabelsAsOrRequestOnlyReturnsTenant() {
         final Map<String, String> labels = new HashMap<>();
         labels.put("os", "DARWIN");
--- a/src/test/java/com/rackspace/salus/resource_management/ResourceManagementTest.java
+++ b/src/test/java/com/rackspace/salus/resource_management/ResourceManagementTest.java
@@ -663,6 +663,50 @@ public class ResourceManagementTest {
     }
 
     @Test
+    public void testMatchResourceWithNoLabelsAsOrRequest() {
+        final Map<String, String> labels = new HashMap<>();
+        labels.put("os", "DARWIN");
+        labels.put("env", "test");
+
+        final Map<String, String> matchLabels = new HashMap<>();
+        labels.put("os", "DARWIN");
+
+        ResourceCreate create = podamFactory.manufacturePojo(ResourceCreate.class);
+        create.setLabels(Collections.emptyMap());
+        String tenantId = RandomStringUtils.randomAlphanumeric(10);
+        resourceManagement.createResource(tenantId, create);
+        entityManager.flush();
+
+        Page<Resource> resources = resourceManagement.getResourcesFromLabels(matchLabels, tenantId, LabelSelectorMethod.OR, Pageable.unpaged());
+        assertEquals(1L, resources.getTotalElements()); //make sure we only returned the one value
+        assertEquals(tenantId, resources.getContent().get(0).getTenantId());
+        assertEquals(create.getResourceId(), resources.getContent().get(0).getResourceId());
+        assertEquals(Collections.emptyMap(), resources.getContent().get(0).getLabels());
+    }
+
+    public void testMatchResourceWithNoLabelsAsOrRequestOnlyReturnsTenant() {
+        final Map<String, String> labels = new HashMap<>();
+        labels.put("os", "DARWIN");
+        labels.put("env", "test");
+
+        final Map<String, String> matchLabels = new HashMap<>();
+        labels.put("os", "DARWIN");
+
+        ResourceCreate create = podamFactory.manufacturePojo(ResourceCreate.class);
+        create.setLabels(Collections.emptyMap());
+        String tenantId = RandomStringUtils.randomAlphanumeric(10);
+        resourceManagement.createResource(tenantId, create);
+        resourceManagement.createResource(RandomStringUtils.randomAlphanumeric(10), create);
+        entityManager.flush();
+
+        Page<Resource> resources = resourceManagement.getResourcesFromLabels(matchLabels, tenantId, LabelSelectorMethod.OR, Pageable.unpaged());
+        assertEquals(1L, resources.getTotalElements()); //make sure we only returned the one value
+        assertEquals(tenantId, resources.getContent().get(0).getTenantId());
+        assertEquals(create.getResourceId(), resources.getContent().get(0).getResourceId());
+        assertEquals(labels, resources.getContent().get(0).getLabels());
+    }
+
+    @Test
     public void testFailedMatchResourceWithMultipleLabels() {
         final Map<String, String> resourceLabels = new HashMap<>();
         resourceLabels.put("os", "DARWIN");
--- a/src/test/java/com/rackspace/salus/resource_management/ResourceManagementTest.java
+++ b/src/test/java/com/rackspace/salus/resource_management/ResourceManagementTest.java
@@ -662,24 +662,35 @@ public class ResourceManagementTest {
         assertEquals(labels, resources.getContent().get(0).getLabels());
     }
 
+    /**
+     * Make sure that when supplied with a emptyMap collection that we return resources with no labels
+     * as well as any resources for that tenant
+     */
     @Test
-    public void testMatchResourceWithNoLabelsAsOrRequest() {
-        final Map<String, String> labels = new HashMap<>();
-        labels.put("os", "DARWIN");
-        labels.put("env", "test");
-
-        final Map<String, String> matchLabels = new HashMap<>();
-        labels.put("os", "DARWIN");
+    public void testMatchResourcesWithEmptyLabels() {
 
         ResourceCreate create = podamFactory.manufacturePojo(ResourceCreate.class);
         create.setLabels(Collections.emptyMap());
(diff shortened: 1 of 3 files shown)


src/test/java/com/rackspace/salus/resource_management/ResourceManagementTest.java
```diff
@@ -662,6 +662,50 @@ public void testMatchResourceWithSingleOrLabel() {
...
+        labels.put("os", "DARWIN");
+
+        ResourceCreate create = podamFactory.manufacturePojo(ResourceCreate.class);
+        create.setLabels(Collections.emptyMap());
+        String tenantId = RandomStringUtils.randomAlphanumeric(10);
+        resourceManagement.createResource(tenantId, create);
+        entityManager.flush();
+
+        Page<Resource> resources = resourceManagement.getResourcesFromLabels(matchLabels, tenantId, LabelSelectorMethod.OR, Pageable.unpaged());
+        assertEquals(1L, resources.getTotalElements()); //make sure we only returned the one value
```
itzg: A resource with empty labels does not work the same as a monitor with an empty label selector. Since `matchLabels` is {os=DARWIN}, then only resources that have at least that label should be returned. Thinking through this, I realized that the only way a resource with empty labels can be matched is by a monitor with an empty label selector (`matchLabels` in this code here).

src/test/java/com/rackspace/salus/resource_management/ResourceManagementTest.java
```diff
@@ -662,6 +662,50 @@ public void testMatchResourceWithSingleOrLabel() {
...
     }
 
+    @Test
+    public void testMatchResourceWithNoLabelsAsOrRequest() {
+        final Map<String, String> labels = new HashMap<>();
+        labels.put("os", "DARWIN");
+        labels.put("env", "test");
+
+        final Map<String, String> matchLabels = new HashMap<>();
+        labels.put("os", "DARWIN");
```
jjbuchan: You add all these entries to `labels` but `labels` is never used.

src/test/java/com/rackspace/salus/resource_management/ResourceManagementTest.java
```diff
@@ -662,6 +662,50 @@ public void testMatchResourceWithSingleOrLabel() {
...
+        assertEquals(Collections.emptyMap(), resources.getContent().get(0).getLabels());
+    }
+
+    public void testMatchResourceWithNoLabelsAsOrRequestOnlyReturnsTenant() {
+        final Map<String, String> labels = new HashMap<>();
+        labels.put("os", "DARWIN");
+        labels.put("env", "test");
+
+        final Map<String, String> matchLabels = new HashMap<>();
+        labels.put("os", "DARWIN");
```
jjbuchan: This line isn't needed.

src/test/java/com/rackspace/salus/resource_management/ResourceManagementTest.java
```diff
@@ -662,6 +662,50 @@ public void testMatchResourceWithSingleOrLabel() {
...
+        entityManager.flush();
+
+        Page<Resource> resources = resourceManagement.getResourcesFromLabels(matchLabels, tenantId, LabelSelectorMethod.OR, Pageable.unpaged());
+        assertEquals(1L, resources.getTotalElements()); //make sure we only returned the one value
+        assertEquals(tenantId, resources.getContent().get(0).getTenantId());
+        assertEquals(create.getResourceId(), resources.getContent().get(0).getResourceId());
+        assertEquals(Collections.emptyMap(), resources.getContent().get(0).getLabels());
+    }
+
+    public void testMatchResourceWithNoLabelsAsOrRequestOnlyReturnsTenant() {
```
jjbuchan: These tests names are kinda confusion to me.  You might want to add a little docs to show what these are testing vs. trying to fit it all in to the name.

src/test/java/com/rackspace/salus/resource_management/ResourceManagementTest.java
harageth: You are correct the matchLabels aren't used but the other labels are needed to pass into the findResourcesFromLabels function. 

And as you pointed out this could be irrelevant due to Geoffs comment.

jjbuchan: `matchLabels` is passed into that method.  `labels` does not get used as far as i can see.  But then it seems like you could just pass in `Collections.emptyMap()` instead of creating `matchLabels` if you were to keep this test.

src/test/java/com/rackspace/salus/resource_management/ResourceManagementTest.java
harageth: Right. I always think through this side incorrectly. I don't think I saw a test where we try and match against passing in an empty collection of labels.

src/test/java/com/rackspace/salus/resource_management/ResourceManagementTest.java
itzg: Actually I missed that this is doing a put on `label` and not `matchLabels`, so you're test is actually/accidentally doing what I commented about.

src/test/java/com/rackspace/salus/resource_management/ResourceManagementTest.java
itzg: ...so this comment is moot I think since `matchLabels` is empty and "all" the resources are correctly getting returned.

src/test/java/com/rackspace/salus/resource_management/ResourceManagementTest.java
harageth: oh wow I totally missed each time that I read this that it was on labels too..

jjbuchan: No one ever listens to me!

itzg: @jjbuchan I was listening to you :) ...just worded my follow-up to your comment poorly

harageth: If it makes you feel better I am fixing it based on your response

GeorgeJahad: yeh, i'm probably the only one who doesn't listen to you

src/test/java/com/rackspace/salus/resource_management/ResourceManagementTest.java
```diff
@@ -662,24 +662,35 @@ public void testMatchResourceWithSingleOrLabel() {
...
+
+
         entityManager.flush();
 
-        Page<Resource> resources = resourceManagement.getResourcesFromLabels(matchLabels, tenantId, LabelSelectorMethod.OR, Pageable.unpaged());
-        assertEquals(1L, resources.getTotalElements()); //make sure we only returned the one value
+        Page<Resource> resources = resourceManagement.getResourcesFromLabels(Collections.emptyMap(), tenantId, LabelSelectorMethod.OR, Pageable.unpaged());
+        assertEquals(2L, resources.getTotalElements()); //make sure we only returned the one value
         assertEquals(tenantId, resources.getContent().get(0).getTenantId());
+        //assertThat(resources.get()., containsInAnyOrder(create.getResourceId(), create2.getResourceId()));
```
itzg: Can be deleted

src/test/java/com/rackspace/salus/resource_management/ResourceManagementTest.java
jjbuchan: Still not sure if this is supposed to be removed or be for `matchLabels`?

src/test/java/com/rackspace/salus/resource_management/ResourceManagementTest.java
jjbuchan: Bump

src/test/java/com/rackspace/salus/resource_management/ResourceManagementTest.java
harageth: It needs changed. I could have sworn I had pushed out an update for this but appears not... getting it fixed up right now.

src/test/java/com/rackspace/salus/resource_management/ResourceManagementTest.java
```diff
@@ -662,6 +662,60 @@ public void testMatchResourceWithSingleOrLabel() {
...
+
+        ResourceCreate create2 = podamFactory.manufacturePojo(ResourceCreate.class);
+        create.setLabels(labels);
+        resourceManagement.createResource(tenantId, create2);
+
+
+        entityManager.flush();
+
+        Page<Resource> resources = resourceManagement.getResourcesFromLabels(Collections.emptyMap(), tenantId, LabelSelectorMethod.OR, Pageable.unpaged());
+        assertEquals(2L, resources.getTotalElements()); //make sure we only returned the one value
```
GeorgeJahad: The comment say one, and the assert says 2?

src/test/java/com/rackspace/salus/resource_management/ResourceManagementTest.java
```diff
@@ -662,6 +662,60 @@ public void testMatchResourceWithSingleOrLabel() {
...
+        assertEquals(tenantId, resources.getContent().get(0).getTenantId());
+        assertEquals(create.getResourceId(), resources.getContent().get(0).getResourceId());
+        assertEquals(Collections.emptyMap(), resources.getContent().get(0).getLabels());
```
GeorgeJahad: since it looks like it is returning two, is there any reason why the test isn't checking both?

PR #70, opened 2020-02-10
Enable role based json views

jjbuchan:
# What

Enable role based json views instead of endpoint driven views.

# How

1. Add `@EnableRoleBasedJsonViews` to app
   * https://github.com/racker/salus-common/pull/49
1. Remove `@JsonView` annotations on api controllers
1. Add default role mappings to `application-dev.yml`

# Why

See https://github.com/racker/salus-common/pull/49


src/test/java/com/rackspace/salus/resource_management/web/controller/ResourceApiControllerTest.java
```diff
@@ -96,8 +99,9 @@
   @Autowired
   ObjectMapper objectMapper;
 
+  @WithMockUser(roles = "CUSTOMER")
```
itzg: It's so nice that spring security test provides these.

jjbuchan: Yeah.  Took me a little while to find this method but really helped keep things simple.

src/main/java/com/rackspace/salus/resource_management/web/model/ResourceDTO.java
```diff
@@ -13,8 +13,7 @@
 public class ResourceDTO {
   @JsonView(View.Admin.class)
   Long id;
-  //@JsonView(View.Public.class) // This will be changed to Internal or Admin when we switch to role based views.
-  // Disabling this until we have a way for our internal requests to bypass that view (this might need new endpoints)
+  @JsonView(View.Internal.class)
```
harageth: Doesn't @FarajiA need to see the tenantId to properly make requests? I guess he probably has it from when he logs in and starts a session for injecting making this redundant. But it also begs the question of why bother limiting this?

jjbuchan: He doesn't need it at all.  This is what's coming back in the resource response.

This field is redundant for almost any request which is why I excluded it.  Right now the only real use-case for including it is for our own cross-service requests since they depend on it being there.

I set it as internal so that if a racker is troubleshooting something using the api the payload will clearly show what tenant they are working with just in case they happened to specify the wrong request.  i.e. there's a tiny chance it might help someone realize they're looking at the wrong thing.  This is probably more likely for those that have service-admin credentials and will be able to query any tenant without impersonation.

PR #72, opened 2020-02-21
added custom json deserializer and updated tests

harageth:
# Resolves

https://jira.rax.io/browse/SALUS-694

# What

It is wanting to lowercase the resourceId's that come in so that it is obvious to users that resourceId's are case insensitive.

# How

It adds a customer JsonDeserializer on the resourceId field that lowercases everything

## How to test

Run the unit tests

# Why

Having this happen at JsonDeserialization time is good because it means that it happens before we get our hand on the POJO itself

### DIFF
--- a/src/main/java/com/rackspace/salus/resource_management/web/model/ResourceCreate.java
+++ b/src/main/java/com/rackspace/salus/resource_management/web/model/ResourceCreate.java
@@ -16,6 +16,8 @@
 
 package com.rackspace.salus.resource_management.web.model;
 
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.rackspace.salus.telemetry.translators.JsonToLowerCaseDeserializer;
 import java.io.Serializable;
 import java.util.Map;
 import javax.validation.constraints.NotNull;
@@ -28,6 +30,7 @@ import javax.validation.constraints.NotBlank;
 public class ResourceCreate implements Serializable {
     @Pattern(regexp="[A-Za-z0-9:-]+")
     @NotBlank
+    @JsonDeserialize(using = JsonToLowerCaseDeserializer.class)
     String resourceId;
 
     Map<String,String> labels;
--- a/src/test/java/com/rackspace/salus/resource_management/web/controller/ResourceApiControllerTest.java
+++ b/src/test/java/com/rackspace/salus/resource_management/web/controller/ResourceApiControllerTest.java
@@ -298,6 +298,7 @@ public class ResourceApiControllerTest {
         .andExpect(content()
             .contentTypeCompatibleWith(MediaType.APPLICATION_JSON));
 
+    create.setResourceId(create.getResourceId().toLowerCase());
     verify(resourceManagement).createResource(tenantId, create);
     verifyNoMoreInteractions(resourceManagement);
   }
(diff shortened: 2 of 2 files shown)


src/main/java/com/rackspace/salus/resource_management/web/model/ResourceCreate.java
```diff
@@ -28,6 +30,7 @@
 public class ResourceCreate implements Serializable {
     @Pattern(regexp="[A-Za-z0-9:-]+")
     @NotBlank
+    @JsonDeserialize(using = JsonToLowerCaseDeserializer.class)
     String resourceId;
```
itzg: @jjbuchan @GeorgeJahad, I wanted to gather your thoughts since I'm on the fence about whether the resource ID given by the user should be altered in order to address case insensitive matching via etcd and `Objects.equals`. For example, if a user created a resource with "MyCustomResourceId" then the entity stored and the returned object will turn that into "mycustomresourceid".

GeorgeJahad: it seems like if we are only going to allow lower case resourceid's then trying to create a resource like "MyCustomResourceId" should return an error.

jjbuchan: Without having read the pr yet or thought about all the problems my preference is that we would store (and return) the user provided value but treat it as case-insensitive.

1. the user should always see the same value they provided
1. failures should occur if they try to create mycustomresourceid when MyCustomResourceId already exists
1. other than that create validation our services should treat MyCustomResourceId and mycustomresourceid as the same thing

harageth: The more I think about it the more I agree with James. 

Per James suggestions this means just changing the collation in the database isn't an option. If we change that to a binary format then we lose case insensitive storage in the db. 

We don't want to do what I have done in this PR in altering the users inputs, which everyone seems to agree on (I didn't feel super comfortable about it but at the time I wasn't positive of a better way to do it)

And to be frank I don't want to go through the entire codebase to make sure all of the comparisons are correct. That is tedious and a recipe for bugs. 

I think that leaves the only option of extending String into a `CaseInsensitiveString` class and overriding the comparison functions to do case insensitive comparisons. This guarantees that comparison will be done correctly whenever it is done on these objects without the chance of introducing bugs and still allows the user input to be stored as is in the database. We can still interact with them as if they are Strings and shouldn't have any storage issues either.

jjbuchan: Again, I've not read the PR and haven't thought about all the problem... but from what I remember the main point of this issue is that we were already treating it case-insensitive in most apps but etcd was not?  So if we just make etcd less strict (or the interactions with etcd more consistent) maybe that eliminated the problem?

I can look more at this on Monday in case that doesn't make sense.

harageth: Yeah I think I misunderstood the original intent of the ticket. This makes it a lot more clear to me. I think my idea about extending String is still something that will need to happen, but the other half that I didn't understand was relaxing etcd storage validation to being case insensitive.

PR #79, opened 2020-04-23
Adam9500/salus 836 update resource details

harageth:
# Resolves

SALUS-836

# What

Add the envoyId into the ResourceDTO response by grabbing it from etcd. It also adds access controls over the returned values so that we can control which types of users can see the values.

# How

We pull in the etcd-adapter library and use the connections there to interact with etcd and get the envoy information.

We can't remove the `associatedWithEnvoy` flag from the Resource Object until we add a full connection history. Without that we can't determine in ResourceManagement whether the envoy connection event was a reconnection or not. 

## How to test

The unit tests should be updated

# Why

I had briefly considered adding endpoints to the ambassador so that we could just ask it for the envoy information. But I was hesitant to add anything like that to a service that didn't have it yet and we already had the library to connect and pull information from etcd anyway.

# TODO

Open the Monitor Management PR

### DIFF
--- a/src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
+++ b/src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
@@ -482,7 +482,7 @@ public class ResourceManagement {
         (resultSet, rowIndex) -> resultSet.getLong(1)
     );
 
-    return resourceRepository.findAllByResourceIdIn(resourceIds, page);
+    return resourceRepository.findByIdIn(resourceIds, page);
 
   }
 
--- a/src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
+++ b/src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
@@ -145,7 +145,7 @@ public class ResourceManagement {
   public ResourceDTO getResourceDTO(String tenantId, String resourceId) {
     Optional<Resource> resource = resourceRepository.findByTenantIdAndResourceId(tenantId, resourceId);
 
-    if(resource.get() == null) {
+    if(!resource.isPresent()) {
       throw new NotFoundException(String.format("No resource found for %s on tenant %s",
           resourceId, tenantId));
     }
@@ -161,7 +161,7 @@ public class ResourceManagement {
   public Page<ResourceDTO> getAllResourceDTOs(Pageable page) {
     final List<ResourceDTO> values = new LinkedList();
     resourceRepository.findAll().forEach(r -> {
-      values.add(new ResourceDTO(r,null)); //envoyResourceManagement.getOne(r.getTenantId(), r.getResourceId()).join().getEnvoyId()));
+      values.add(getResourceDTOFromResource(r));
     });
     return new PageImpl(values, page, values.size());
   }
--- a/src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
+++ b/src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
@@ -159,11 +159,12 @@ public class ResourceManagement {
    * @return The resourceDTOs found that match the page criteria.
    */
   public Page<ResourceDTO> getAllResourceDTOs(Pageable page) {
-    final List<ResourceDTO> values = new LinkedList();
-    resourceRepository.findAll().forEach(r -> {
-      values.add(getResourceDTOFromResource(r));
+    final List<ResourceDTO> resourceDTOs = new LinkedList();
+    Page<Resource> resources = resourceRepository.findAll(page);
+    resources.forEach(r -> {
+      resourceDTOs.add(getResourceDTOFromResource(r));
     });
-    return new PageImpl(values, page, values.size());
+    return new PageImpl(resourceDTOs, page, resources.getTotalElements());
   }
 
   /**
@@ -173,17 +174,16 @@ public class ResourceManagement {
    * @return The resources found for the tenant that match the page criteria.
    */
   public Page<ResourceDTO> getResourceDTOs(String tenantId, Pageable page) {
-    final List<ResourceDTO> values = new LinkedList();
-    resourceRepository.findAllByTenantId(tenantId).forEach(r -> {
-      values.add(getResourceDTOFromResource(r));
+    final List<ResourceDTO> resourceDTOs = new LinkedList();
+    Page<Resource> resources = resourceRepository.findAllByTenantId(tenantId, page);
+
+    resources.forEach(r -> {
+      resourceDTOs.add(getResourceDTOFromResource(r));
     });
 
-    return new PageImpl(values, page, values.size());
+    return new PageImpl(resourceDTOs, page, resources.getTotalElements());
   }
 
-  public List<Resource> getAllTenantResources(String tenantId) {
-    return resourceRepository.findAllByTenantId(tenantId);
-  }
   /**
    * Get all resources where the presence monitoring field matches the parameter provided.
    * @param presenceMonitoringEnabled Whether presence monitoring is enabled or not.
@@ -192,8 +192,7 @@ public class ResourceManagement {
   public Stream<ResourceDTO> getResources(boolean presenceMonitoringEnabled) {
     return resourceRepository.findAllByPresenceMonitoringEnabled(presenceMonitoringEnabled)
         .stream()
-        .map(r -> getResourceDTOFromResource(r))
-        .collect(Collectors.toList()).stream();
+        .map(this::getResourceDTOFromResource);
   }
 
   /**
@@ -439,7 +438,7 @@ public class ResourceManagement {
     pagedResources.get().forEach(r -> {
       values.add(getResourceDTOFromResource(r));
     });
-    return (Page<ResourceDTO>) getPagedResults(values, page);
+    return new PageImpl(values, page, pagedResources.getTotalElements());
   }
 
   /**
@@ -451,7 +450,7 @@ public class ResourceManagement {
    */
   public Page<Resource> getResourcesFromLabels(Map<String, String> labels, String tenantId, LabelSelectorMethod logicalOperation, Pageable page) {
     if(labels == null || labels.isEmpty()) {
-      return (Page<Resource>) getPagedResults(getAllTenantResources(tenantId), page);
+      return resourceRepository.findAllByTenantId(tenantId, page);
     }
 
     MapSqlParameterSource paramSource = new MapSqlParameterSource();
@@ -483,12 +482,8 @@ public class ResourceManagement {
         (resultSet, rowIndex) -> resultSet.getLong(1)
     );
 
-    final List<Resource> resources = new ArrayList<>();
-    for (Resource resource : resourceRepository.findAllById(resourceIds)) {
-      resources.add(resource);
-    }
+    return resourceRepository.findAllByResourceIdIn(resourceIds, page);
 
-    return (Page<Resource>) getPagedResults(resources, page);
   }
 
 
@@ -525,32 +520,6 @@ public class ResourceManagement {
         .flatMap(map -> map.keySet().stream())
         .distinct()
         .collect(Collectors.toList());
-  }
-
-  /**
-   * Converts a list of resources into a Page containing only the requested elements.
-   * @param resources The full list of resources found.
-   * @param page The page details that were requested.
-   * @return A Page object containing only the resources that were requested.
-   */
-  @SuppressWarnings("Duplicates")
-  private Page<?> getPagedResults(List<?> resources, Pageable page) {
-    if (page.isPaged()) {
-      int start = page.getPageSize() * page.getPageNumber();
-      int end = start + page.getPageSize();
-      if (end > resources.size()) {
-        end = resources.size();
-      }
-      List<?> results;
-      try {
-        results = resources.subList(start, end);
-      } catch (IndexOutOfBoundsException | IllegalArgumentException e) {
-        results = Collections.emptyList();
-      }
-      return new PageImpl<>(results, page, resources.size());
-    } else {
-      return new PageImpl<>(resources, page, resources.size());
-    }
   }
 
   private ResourceDTO getResourceDTOFromResource(Resource resource) {
--- a/src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
+++ b/src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
@@ -159,12 +159,8 @@ public class ResourceManagement {
    * @return The resourceDTOs found that match the page criteria.
    */
   public Page<ResourceDTO> getAllResourceDTOs(Pageable page) {
-    final List<ResourceDTO> resourceDTOs = new LinkedList();
-    Page<Resource> resources = resourceRepository.findAll(page);
-    resources.forEach(r -> {
-      resourceDTOs.add(getResourceDTOFromResource(r));
-    });
-    return new PageImpl(resourceDTOs, page, resources.getTotalElements());
+    return resourceRepository.findAll(page)
+        .map(this::getResourceDTOFromResource);
   }
 
   /**
@@ -174,14 +170,8 @@ public class ResourceManagement {
    * @return The resources found for the tenant that match the page criteria.
    */
   public Page<ResourceDTO> getResourceDTOs(String tenantId, Pageable page) {
-    final List<ResourceDTO> resourceDTOs = new LinkedList();
-    Page<Resource> resources = resourceRepository.findAllByTenantId(tenantId, page);
-
-    resources.forEach(r -> {
-      resourceDTOs.add(getResourceDTOFromResource(r));
-    });
-
-    return new PageImpl(resourceDTOs, page, resources.getTotalElements());
+    return resourceRepository.findAllByTenantId(tenantId, page)
+        .map(this::getResourceDTOFromResource);
   }
 
   /**
@@ -433,12 +423,8 @@ public class ResourceManagement {
   }
 
   public Page<ResourceDTO> getResourceDTOsFromLabels(Map<String, String> labels, String tenantId, LabelSelectorMethod logicalOperation, Pageable page) {
-    Page<Resource> pagedResources = getResourcesFromLabels(labels, tenantId, logicalOperation, page);
-    final List<ResourceDTO> values = new LinkedList();
-    pagedResources.get().forEach(r -> {
-      values.add(getResourceDTOFromResource(r));
-    });
-    return new PageImpl(values, page, pagedResources.getTotalElements());
+    return getResourcesFromLabels(labels, tenantId, logicalOperation, page)
+      .map(this::getResourceDTOFromResource);
   }
 
   /**
--- a/src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
+++ b/src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
@@ -21,14 +21,17 @@ import static com.rackspace.salus.telemetry.model.LabelNamespaces.labelHasNamesp
 import com.rackspace.salus.common.util.SpringResourceUtils;
 import com.rackspace.salus.resource_management.config.ResourceManagementProperties;
 import com.rackspace.salus.resource_management.web.model.ResourceCreate;
+import com.rackspace.salus.resource_management.web.model.ResourceDTO;
 import com.rackspace.salus.resource_management.web.model.ResourceUpdate;
 import com.rackspace.salus.telemetry.entities.Resource;
 import com.rackspace.salus.telemetry.errors.AlreadyExistsException;
+import com.rackspace.salus.telemetry.etcd.services.EnvoyResourceManagement;
 import com.rackspace.salus.telemetry.messaging.AttachEvent;
 import com.rackspace.salus.telemetry.messaging.ResourceEvent;
 import com.rackspace.salus.telemetry.model.LabelNamespaces;
 import com.rackspace.salus.telemetry.model.LabelSelectorMethod;
 import com.rackspace.salus.telemetry.model.NotFoundException;
+import com.rackspace.salus.telemetry.model.ResourceInfo;
 import com.rackspace.salus.telemetry.repositories.ResourceRepository;
 import java.io.IOException;
 import java.util.ArrayList;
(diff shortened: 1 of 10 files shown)


src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
```diff
@@ -133,23 +142,43 @@ public boolean exists(String tenantId, String resourceId) {
...
+
   /**
    * Get a selection of resource objects across all accounts.
    * @param page The slice of results to be returned.
-   * @return The resources found that match the page criteria.
+   * @return The resourceDTOs found that match the page criteria.
    */
-  public Page<Resource> getAllResources(Pageable page) {
-    return resourceRepository.findAll(page);
+  public Page<ResourceDTO> getAllResourceDTOs(Pageable page) {
```
itzg, line 161: This `page` request should be provided to the repository method to let the SQL query to the limit+offset. With that `Page<?>` response you can remap the values using [Page.map](https://docs.spring.io/spring-data/data-commons/docs/current/api/org/springframework/data/domain/Page.html#map-java.util.function.Function-)

src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
```diff
@@ -133,23 +142,43 @@ public boolean exists(String tenantId, String resourceId) {
...
   /**
-   * Same as {@link #getAllResources(Pageable page) getAllResources} except restricted to a single tenant.
+   * Same as {@link #getAllResourceDTOs(Pageable page) getAllResourceDTOs} except restricted to a single tenant.
    * @param tenantId The tenant to select resources from.
    * @param page The slice of results to be returned.
    * @return The resources found for the tenant that match the page criteria.
    */
-  public Page<Resource> getResources(String tenantId, Pageable page) {
-    return resourceRepository.findAllByTenantId(tenantId, page);
+  public Page<ResourceDTO> getResourceDTOs(String tenantId, Pageable page) {
```
itzg, line 175: Same thing here with `page` not being used

src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
```diff
@@ -160,8 +189,11 @@ public boolean exists(String tenantId, String resourceId) {
    * @param presenceMonitoringEnabled Whether presence monitoring is enabled or not.
    * @return Stream of resources.
    */
-  public Stream<Resource> getResources(boolean presenceMonitoringEnabled) {
-    return resourceRepository.findAllByPresenceMonitoringEnabled(presenceMonitoringEnabled).stream();
+  public Stream<ResourceDTO> getResources(boolean presenceMonitoringEnabled) {
+    return resourceRepository.findAllByPresenceMonitoringEnabled(presenceMonitoringEnabled)
+        .stream()
+        .map(r -> getResourceDTOFromResource(r))
```
itzg, line 195: IntelliJ might be suggesting to you
```suggestion
        .map(this::getResourceDTOFromResource)
```

src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
```diff
@@ -160,8 +189,11 @@ public boolean exists(String tenantId, String resourceId) {
...
+        .collect(Collectors.toList()).stream();
```
itzg, line 196: I think if you leave off this call the result of the previous line's `.map` is a `Stream`.

src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
```diff
@@ -401,6 +433,15 @@ private void removePresenceMonitoring(String tenantId, String resourceId) {
     saveAndPublishResource(resource, false, null);
   }
 
+  public Page<ResourceDTO> getResourceDTOsFromLabels(Map<String, String> labels, String tenantId, LabelSelectorMethod logicalOperation, Pageable page) {
+    Page<Resource> pagedResources = getResourcesFromLabels(labels, tenantId, logicalOperation, page);
+    final List<ResourceDTO> values = new LinkedList();
+    pagedResources.get().forEach(r -> {
+      values.add(getResourceDTOFromResource(r));
+    });
+    return (Page<ResourceDTO>) getPagedResults(values, page);
```
itzg, line 442: Use `Page.map`

src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
```diff
@@ -410,7 +451,7 @@ private void removePresenceMonitoring(String tenantId, String resourceId) {
    */
   public Page<Resource> getResourcesFromLabels(Map<String, String> labels, String tenantId, LabelSelectorMethod logicalOperation, Pageable page) {
     if(labels == null || labels.isEmpty()) {
-      return getPagedResults(getAllTenantResources(tenantId), page);
+      return (Page<Resource>) getPagedResults(getAllTenantResources(tenantId), page);
```
itzg, line 454: This cast makes me nervous. I'm not quite sure why you need getPageResults in the first place, but have it be a generic, parameterized method to avoid the `<?>` usage:

https://docs.oracle.com/javase/tutorial/extra/generics/methods.html

src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
```diff
@@ -493,14 +534,14 @@ private void removePresenceMonitoring(String tenantId, String resourceId) {
    * @return A Page object containing only the resources that were requested.
    */
   @SuppressWarnings("Duplicates")
-  private Page<Resource> getPagedResults(List<Resource> resources, Pageable page) {
+  private Page<?> getPagedResults(List<?> resources, Pageable page) {
```
itzg, line 537: This method seems unnecessary if the SQL offset/limit is used instead by passing `page` to repository methods.

src/main/java/com/rackspace/salus/resource_management/web/model/ResourceDTO.java
```diff
@@ -16,23 +16,27 @@
   Long id;
   @JsonView(View.Internal.class)
   String tenantId;
+  @JsonView(View.Admin.class)
+  String envoyId;
+  @JsonView(View.Internal.class)
+  boolean associatedWithEnvoy;
```
itzg, line 22: I would have thought users via Intelligence would want to see these.

src/test/java/com/rackspace/salus/resource_management/ResourceManagementTest.java
```diff
@@ -157,7 +176,15 @@ public void testCreateNewResource() {
...
-        Resource returned = resourceManagement.createResource(tenantId, create);
+        ResourceInfo info = new ResourceInfo()
+            .setEnvoyId("e-1")
+            .setTenantId(tenantId)
+            .setResourceId(resourceId);
+
+        when(envoyResourceManagement.getOne(any(), any()))
+            .thenReturn(CompletableFuture.completedFuture(info));
+
+        ResourceDTO returned = resourceManagement.createResource(tenantId, create);
```
itzg, line 187: Would be good to assert the envoy field was set as expected

src/test/java/com/rackspace/salus/resource_management/ResourceManagementTest.java
```diff
@@ -178,17 +205,23 @@ public void testGetAll() {
...
+        Page<ResourceDTO> result = resourceManagement.getAllResourceDTOs(page);
 
         assertThat(result.getTotalElements(), equalTo(1L));
 
         // There is already one resource created as default
         createResources(totalResources - 1);
 
         page = PageRequest.of(0, 10);
-        result = resourceManagement.getAllResources(page);
+        result = resourceManagement.getAllResourceDTOs(page);
```
itzg, line 224: I'd say assert the envoy is set in at least one entry, but previous test didn't assert any of the fields.

src/test/java/com/rackspace/salus/resource_management/ResourceManagementTest.java
```diff
@@ -483,6 +545,8 @@ public void testUpdateExistingResource() {
             return;
         }
 
+        resource = resourceManagement.getAllResourceDTOs(PageRequest.of(0, 1)).getContent().get(0);
```
itzg, line 548: Assert envoyId

src/test/java/com/rackspace/salus/resource_management/ResourceManagementTest.java
```diff
@@ -528,7 +600,7 @@ public void testUpdateExistingResourceWithMetadata() {
         assertThat(newResource.getLabels(), equalTo(expectedLabels));
         assertThat(newResource.getMetadata(), equalTo(resource.getMetadata()));
         assertThat(newResource.getId(), equalTo(resource.getId()));
-        assertThat(newResource.getPresenceMonitoringEnabled(), equalTo(presenceMonitoring));
+        //assertThat(newResource.getPresenceMonitoringEnabled(), equalTo(presenceMonitoring));
```
itzg, line 603: Needs to be reverted?

src/test/java/com/rackspace/salus/resource_management/web/controller/ResourceApiControllerTest.java
```diff
@@ -203,47 +207,52 @@ public void testGetByResourceIdAsEmployee() throws Exception {
...
 
-    mockMvc.perform(get(
+    ResultActions value = mockMvc.perform(get(
         "/api/tenant/{tenantId}/resources/{resourceId}",
         "t-1", "r-1"
-    ).accept(MediaType.APPLICATION_JSON))
+    ).accept(MediaType.APPLICATION_JSON));
+
+
+        value
```
itzg, line 237: Could be chained. Besides `value` is a odd variable name here. Would have expected `actions` or `resultActions`.

harageth: oh and I see I left a large amount of whitespace in there. Thought I had cleaned that up.

src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
harageth: Ahhh I see what I originally did wrong. I converted the values to return in the foreach loop and then used the values.size() rather than the paged result `.getTotalElements()`

src/main/java/com/rackspace/salus/resource_management/web/model/ResourceDTO.java
harageth: This was a recommendation by @jjbuchan. His point was that customers shouldn't need to know these sorts of specifics about the system.

itzg: Yeah, I'm on the fence since I agree that the specific envoy ID probably isn't useful, but I could see customers wanting to see "envoy attached" kind of status...however, while grabbing this screenshot from Intelligence, I see that we report agent ID there:

image

src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
harageth: So the one function that did need it I just altered in the Repository and cleaned up the rest of it. I'll push that PR out shortly.

src/main/java/com/rackspace/salus/resource_management/web/model/ResourceDTO.java
jjbuchan: I haven't reviewed this PR yet, but if `associatedWithEnvoy` still acts like it did before and could be `true` even if not connected then I don't think this should be exposed.  In that state it can cause confusion.

In another PR we could rename this to something more accurate and create a field on the DTO that is set true if an envoyId is set and false if one is not.

`Agent ID` in Intelligence is the equivalent to `resourceId` for our envoys.  Its the `agent_id` on the entity that is also specified in the agent's config file.  For a cloud server it will be the xen-id.

harageth: Yes it still acts the same way as it did before. I attempted to change it to be off of the envoyId but that still caused the same issue I had hit before where we needed to full history to be able to detect reconnects.

src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
```diff
@@ -159,11 +159,12 @@ public ResourceDTO getResourceDTO(String tenantId, String resourceId) {
...
-    final List<ResourceDTO> values = new LinkedList();
-    resourceRepository.findAll().forEach(r -> {
-      values.add(getResourceDTOFromResource(r));
+    final List<ResourceDTO> resourceDTOs = new LinkedList();
+    Page<Resource> resources = resourceRepository.findAll(page);
+    resources.forEach(r -> {
+      resourceDTOs.add(getResourceDTOFromResource(r));
     });
-    return new PageImpl(values, page, values.size());
+    return new PageImpl(resourceDTOs, page, resources.getTotalElements());
```
itzg: `Page` has a `map` method that is a cleaner way to do this object re-mapping. You should be able to find examples nearby this code. I think it becomes

```
return resourceRepository.findAll(page)
  .map(this::getResourceDTOFromResource);
```

src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
```diff
@@ -173,17 +174,16 @@ public ResourceDTO getResourceDTO(String tenantId, String resourceId) {
...
-      values.add(getResourceDTOFromResource(r));
+    final List<ResourceDTO> resourceDTOs = new LinkedList();
+    Page<Resource> resources = resourceRepository.findAllByTenantId(tenantId, page);
+
+    resources.forEach(r -> {
+      resourceDTOs.add(getResourceDTOFromResource(r));
     });
 
-    return new PageImpl(values, page, values.size());
+    return new PageImpl(resourceDTOs, page, resources.getTotalElements());
```
itzg: Use `Page.map`

src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
```diff
@@ -439,7 +438,7 @@ private void removePresenceMonitoring(String tenantId, String resourceId) {
     pagedResources.get().forEach(r -> {
       values.add(getResourceDTOFromResource(r));
     });
-    return (Page<ResourceDTO>) getPagedResults(values, page);
+    return new PageImpl(values, page, pagedResources.getTotalElements());
```
itzg: Can use `Page.map`

src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
itzg: Not sure why github shows "outdated"...so hopefully this comment is still applicable to latest changes.

harageth: Not sure why it is either. Yes its still applicable. I must have misunderstood your other comments that had suggested this in your first review.

harageth: Updated this so resolving

src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
```diff
@@ -133,35 +142,47 @@ public boolean exists(String tenantId, String resourceId) {
...
    * Get all resources where the presence monitoring field matches the parameter provided.
    * @param presenceMonitoringEnabled Whether presence monitoring is enabled or not.
    * @return Stream of resources.
    */
-  public Stream<Resource> getResources(boolean presenceMonitoringEnabled) {
-    return resourceRepository.findAllByPresenceMonitoringEnabled(presenceMonitoringEnabled).stream();
+  public Stream<ResourceDTO> getResources(boolean presenceMonitoringEnabled) {
+    return resourceRepository.findAllByPresenceMonitoringEnabled(presenceMonitoringEnabled)
+        .stream()
+        .map(this::getResourceDTOFromResource);
```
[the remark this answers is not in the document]
jjbuchan: Oh, was it to use the new constructor and populate envoyId here instead of the controller?

src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
```diff
@@ -133,35 +142,47 @@ public boolean exists(String tenantId, String resourceId) {
...
-    return resourceRepository.findAllByTenantId(tenantId);
-  }
   /**
    * Get all resources where the presence monitoring field matches the parameter provided.
    * @param presenceMonitoringEnabled Whether presence monitoring is enabled or not.
    * @return Stream of resources.
    */
-  public Stream<Resource> getResources(boolean presenceMonitoringEnabled) {
-    return resourceRepository.findAllByPresenceMonitoringEnabled(presenceMonitoringEnabled).stream();
+  public Stream<ResourceDTO> getResources(boolean presenceMonitoringEnabled) {
```
jjbuchan, line 182: The below question was a general one about all methods, but incidentally, I think this method should be different from others.

This `Stream` request is only used internally for presence monitoring. I don't think we should lookup etcd for each of these because it is not required by presence monitoring and just adds more work.

(commenting up here to keep the discussion separate from my other question)

src/main/java/com/rackspace/salus/resource_management/web/model/ResourceDTO.java
```diff
@@ -16,23 +16,27 @@
   Long id;
   @JsonView(View.Internal.class)
   String tenantId;
+  @JsonView(View.Admin.class)
+  String envoyId;
```
jjbuchan, line 20: To clarify my comment below, in a future PR I think we should add a field like so:

```
boolean envoyConnected;
```

It will have a default json view (`View.Public.class`) and will be set to `true` if `envoyId` is not empty.

jjbuchan: ...but I am not doubting the need.  It all depends on whether we want to expose `envoyId` or not.

src/main/java/com/rackspace/salus/resource_management/web/model/ResourceDTO.java
```diff
@@ -16,23 +16,27 @@
...
   String resourceId;
   Map<String,String> labels = Collections.emptyMap();
   Map<String,String> metadata = Collections.emptyMap();
   Boolean presenceMonitoringEnabled;
-  boolean associatedWithEnvoy;
   String createdTimestamp;
   String updatedTimestamp;
 
-  public ResourceDTO(Resource resource) {
+  public ResourceDTO(Resource resource, String envoyId) {
```
jjbuchan, line 30: Part of me thinks keeping the old constructor in addition to this one could be helpful, but part of me likes that this means we're explicitly setting the envoyId or not.

src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
jjbuchan: Saying that, I've not looked at Presence Monitoring.  Maybe if we include it here we can remove some etcd lookups from there.

Maybe that's worth looking into to see if this should be removed from here or if changes should be made over there.

src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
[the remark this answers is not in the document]
harageth: Yes. Specifically because I didn't want any database access happening in the controller.

src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
harageth: That's reasonable. So sounds like the options for this one are either:
1. make this stream response of Resource instead of ResourceDTO
or
2. look at presence monitoring to see if we can reduce the etcd lookups in there.

src/main/java/com/rackspace/salus/resource_management/web/model/ResourceDTO.java
harageth: I think that sounds reasonable. We could even add the `boolean envoyConnected` right now. Since we are looking into etcd that should give us the immediate information of whether the envoy is actually connected. 

Any history is strictly needed for understanding when we have a reconnect happening.

src/main/java/com/rackspace/salus/resource_management/web/model/ResourceDTO.java
harageth: I don't really like the idea of passing ResourceDTO's around half constructed. If there are instances where we don't need the envoyId we should be passing `Resource`'s around instead.

src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
```diff
@@ -133,35 +142,47 @@ public boolean exists(String tenantId, String resourceId) {
...
 
+  public ResourceDTO getResourceDTO(String tenantId, String resourceId) {
+    Optional<Resource> resource = resourceRepository.findByTenantIdAndResourceId(tenantId, resourceId);
+
+    if(!resource.isPresent()) {
+      throw new NotFoundException(String.format("No resource found for %s on tenant %s",
+          resourceId, tenantId));
+    }
+
+    return getResourceDTOFromResource(resource.get());
```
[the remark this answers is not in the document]
harageth: Ahhh yeah that is much improved!

src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
harageth: per conversation with @jjbuchan this should be reverted to returning `Stream<Resource>` and then we should create a ticket to look at presence monitoring to see if we can simplify by returning the DTO and reducing the etcd lookups that presence monitoring needs to do.

src/main/java/com/rackspace/salus/resource_management/web/controller/ResourceApiController.java
```diff
@@ -98,7 +95,7 @@ public SseEmitter getAllWithPresenceMonitoringAsStream() {
     taskExecutor.execute(() -> {
       resourcesWithEnvoys.forEach(r -> {
         try {
-          emitter.send(new ResourceDTO(r));
+          emitter.send(r);
```
jjbuchan, line 98:
```suggestion
          emitter.send(new ResourceDTO(r, null));
```

PR #82, opened 2020-07-02
resource search endpoint

harageth:
# Resolves

https://jira.rax.io/browse/SALUS-901

# What

Make sure that we have functions to search for resources

# How

Adds api endpoints for searching based on resourceId

## How to test

Run the unit tests

# Why

Since the repositories were updated this was straightforward.

# TODO

Open a new JIRA covering database indexing

### DIFF
--- a/src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
+++ b/src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
@@ -514,7 +514,7 @@ public class ResourceManagement {
     return LabelNamespaces.getNamespaces();
   }
 
-  public Page<Resource> resourceSearch(String tenantId, String searchCriteria, Pageable page) {
+  public Page<Resource> getResourcesBySearchString(String tenantId, String searchCriteria, Pageable page) {
     return resourceRepository.findByTenantIdAndResourceIdContaining(tenantId, searchCriteria, page);
   }
 }
--- a/src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
+++ b/src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
@@ -12,6 +12,7 @@
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
  * See the License for the specific language governing permissions and
  * limitations under the License.
+ *
  */
 
 package com.rackspace.salus.resource_management.services;
@@ -511,5 +512,9 @@ public class ResourceManagement {
 
   public Collection<String> getLabelNamespaces() {
     return LabelNamespaces.getNamespaces();
+  }
+
+  public Page<Resource> resourceSearch(String tenantId, String searchCriteria, Pageable page) {
+    return resourceRepository.findByTenantIdAndResourceIdContaining(tenantId, searchCriteria, page);
   }
 }
--- a/src/main/java/com/rackspace/salus/resource_management/web/controller/ResourceApiController.java
+++ b/src/main/java/com/rackspace/salus/resource_management/web/controller/ResourceApiController.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019 Rackspace US, Inc.
+ * Copyright 2020 Rackspace US, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -12,6 +12,7 @@
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
  * See the License for the specific language governing permissions and
  * limitations under the License.
+ *
  */
 
 package com.rackspace.salus.resource_management.web.controller;
@@ -185,5 +186,11 @@ public class ResourceApiController {
   @ApiOperation("Lists the label namespaces that are reserved for use by the system")
   public Collection<String> getLabelNamespaces(@PathVariable String tenantId) {
     return resourceManagement.getLabelNamespaces();
+  }
+
+  @GetMapping("/tenant/{tenantId}/resource-search/{searchCriteria}")
+  @ApiOperation("Lists the label namespaces that are reserved for use by the system")
+  public PagedContent<Resource> getResourcesSearch(@PathVariable String tenantId, @PathVariable String searchCriteria, Pageable page) {
+    return PagedContent.fromPage(resourceManagement.resourceSearch(tenantId, searchCriteria, page));
   }
 }
--- a/src/main/java/com/rackspace/salus/resource_management/web/controller/ResourceApiController.java
+++ b/src/main/java/com/rackspace/salus/resource_management/web/controller/ResourceApiController.java
@@ -188,8 +188,8 @@ public class ResourceApiController {
     return resourceManagement.getLabelNamespaces();
   }
 
-  @GetMapping("/tenant/{tenantId}/resource-search/{searchCriteria}")
-  @ApiOperation("Lists the label namespaces that are reserved for use by the system")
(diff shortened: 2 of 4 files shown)


src/main/java/com/rackspace/salus/resource_management/web/controller/ResourceApiController.java
```diff
@@ -186,4 +187,10 @@ public void delete(@PathVariable String tenantId,
   public Collection<String> getLabelNamespaces(@PathVariable String tenantId) {
     return resourceManagement.getLabelNamespaces();
   }
+
+  @GetMapping("/tenant/{tenantId}/search/{searchCriteria}")
+  @ApiOperation("Find all resources that match the searchCriteria in the ResourceId")
+  public PagedContent<Resource> getResourcesSearch(@PathVariable String tenantId, @PathVariable String searchCriteria, Pageable page) {
```
itzg, line 193: For a search criteria it seems like it would be better as a query parameter such as:
```suggestion
  @GetMapping("/tenant/{tenantId}/search")
  @ApiOperation("Find all resources that match the searchCriteria in the ResourceId")
  public PagedContent<Resource> getResourcesSearch(@PathVariable String tenantId, @RequestParam("q") String searchCriteria, Pageable page) {
```

src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
```diff
@@ -512,4 +513,8 @@ private ResourceDTO getResourceDTOFromResource(Resource resource) {
   public Collection<String> getLabelNamespaces() {
     return LabelNamespaces.getNamespaces();
   }
+
+  public Page<Resource> resourceSearch(String tenantId, String searchCriteria, Pageable page) {
```
jjbuchan, line 517:
```suggestion
  public Page<Resource> getResourcesBySearchString(String tenantId, String searchCriteria, Pageable page) {
```

src/main/java/com/rackspace/salus/resource_management/web/controller/ResourceApiController.java
jjbuchan, line 193:
```suggestion
  public PagedContent<Resource> getResourcesBySearchString(@PathVariable String tenantId, @PathVariable String searchCriteria, Pageable page) {
```

Maybe the same here too?

src/test/java/com/rackspace/salus/resource_management/ResourceManagementTest.java
```diff
@@ -1219,6 +1220,20 @@ public void testGetTenantResourceMetadataKeys_whenEmpty() {
...
+      persistResource("t-1", "ping", Collections.emptyMap(), Collections.emptyMap());
+      persistResource("t-1", "CPU", Collections.emptyMap(), Collections.emptyMap());
+      persistResource("t-1", "databasingEverything", Collections.emptyMap(), Collections.emptyMap());
+      persistResource("t-2", "ping", Collections.emptyMap(), Collections.emptyMap());
+
+      Pageable page = PageRequest.of(0, 1);
+      Page<Resource> resources = resourceManagement.resourceSearch("t-1", "in", page);
+      assertThat(resources.getTotalElements(), equalTo(2L));
+      assertThat(resources.getTotalPages(), equalTo(2));
+      assertThat(resources.getNumberOfElements(), equalTo(1));
```
jjbuchan, line 1234: I'd go a step further an also assert the resourceId is the one we expect.

harageth: So I did do that. I was a little hesitant to because its unclear to me how JPA is ordering the result set so I am a little concerned that this test will fail randomly because its ordered a different way.

itzg: Hmm, maybe the resource search should include an ordering on resource ID. I'd say to do the same for monitor search, but there's not an equivalent, user-facing field that is guaranteed to be present/sortable.

harageth: On the query for the monitors I added an `ORDER BY monitor_name` which I think is a reasonable user facing field that they would expect to be ordered on.

itzg: Downside is that monitor name is optional; however, ordering by that is still generally a good thing.

jjbuchan: > I added an `ORDER BY monitor_name`

I don't see that anywhere?

But it also looks like things can be ordered on the client side (based on `Sort.by("resourceId").descending()`) so if that's the case we'd want to make sure the ordering we do doesn't override any custom ordering the client does.

jjbuchan: Oh, you mean in your model PR that already was merged?

src/test/java/com/rackspace/salus/resource_management/web/controller/ResourceApiControllerTest.java
```diff
@@ -711,4 +711,28 @@ public void testGetLabelNamespaces() throws Exception {
 
     verifyNoMoreInteractions(resourceManagement);
   }
+
+  @Test
+  public void testSearchResources() throws Exception {
+
+    final List<ResourceDTO> expectedResources = IntStream.range(0, 4)
+        .mapToObj(value -> podamFactory.manufacturePojo(ResourceDTO.class))
+        .collect(Collectors.toList());
```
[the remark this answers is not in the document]
harageth: how does it know to create 5 of them?

harageth:
@jjbuchan https://github.com/racker/salus-telemetry-model/blob/master/src/main/java/com/rackspace/salus/telemetry/repositories/MonitorRepository.java#L88

At the end of this query is where I added the `ORDER BY`

src/test/java/com/rackspace/salus/resource_management/ResourceManagementTest.java
harageth: https://github.com/racker/salus-telemetry-model/blob/master/src/main/java/com/rackspace/salus/telemetry/repositories/MonitorRepository.java#L88

Right here

src/test/java/com/rackspace/salus/resource_management/web/controller/ResourceApiControllerTest.java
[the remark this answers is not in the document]
jjbuchan: That's just the functionality of the library when you tell it to create an ArrayList.

src/test/java/com/rackspace/salus/resource_management/ResourceManagementTest.java
itzg: Oops, that was probably my faulty suggestion. Now I'm thinking that shouldn't be hardcoded there, but rather accepted via the Pageable. Unless did you find the native query wasn't honoring the sort parameter?

harageth: I dont think it was your suggestion. Its a force of habit when writing SQL queries. I will check the ordering from the client shortly.

harageth: And for posterity's sake I should leave a comment here. https://www.baeldung.com/spring-data-jpa-query 

`When the @Query annotation uses native SQL, then it’s not possible to define a Sort.
If we do, we’ll receive an exception`

So I am leaving the ORDER BY in the repository as is. I did add comments inline and to part of the API documentation indicating we were ignoring the sorting.

PR #86, opened 2020-08-03
Adam/delete tenant

harageth:
# Resolves

https://jira.rax.io/browse/SALUS-956

# What

Delete all resources for a particular tenant

# How

Utilizes the new repository function to delete all resources associated with a tenant.

## How to test

Unit tests have been added

# Why

It seems silly to not do this and instead delete everything individually.

### DIFF
--- a/src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
+++ b/src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
@@ -519,17 +519,19 @@ public class ResourceManagement {
         .map(this::getResourceDTOFromResource);
   }
 
-  public void removeAllTenantResources(String tenantId) {
-    List<Resource> resources = resourceRepository.findAllByTenantId(tenantId);
+  public void removeAllTenantResources(String tenantId, boolean sendEvents) {
+
 
     resourceRepository.deleteAllByTenantId(tenantId);
 
-    resources.forEach(resource ->
-        publishResourceEvent(
-            new ResourceEvent()
-                .setTenantId(tenantId)
-                .setResourceId(resource.getResourceId())
-                .setDeleted(true)
-        ));
+    if(sendEvents) {
+      resourceRepository.findAllByTenantId(tenantId).forEach(resource ->
+          publishResourceEvent(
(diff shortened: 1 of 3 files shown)


src/main/java/com/rackspace/salus/resource_management/web/controller/ResourceApiController.java
```diff
@@ -193,4 +193,10 @@ public void delete(@PathVariable String tenantId,
   public PagedContent<ResourceDTO> getResourcesBySearchString(@PathVariable String tenantId, @RequestParam("q") String searchCriteria, Pageable page) {
     return PagedContent.fromPage(resourceManagement.getResourcesBySearchString(tenantId, searchCriteria, page));
   }
+
+  @DeleteMapping("/tenant/{tenantId}/")
```
itzg, line 197: To avoid accidental use via public API, I suggest an admin qualified path that's purposely non-standard like:
```suggestion
  @DeleteMapping("/admin/delete-all/{tenantId}/")
```

src/main/java/com/rackspace/salus/resource_management/services/ResourceManagement.java
```diff
@@ -518,4 +518,8 @@ private ResourceDTO getResourceDTOFromResource(Resource resource) {
     return resourceRepository.findByTenantIdAndResourceIdContaining(tenantId, searchCriteria, page)
         .map(this::getResourceDTOFromResource);
   }
+
+  public void removeAllTenantResources(String tenantId) {
+    resourceRepository.deleteAllByTenantId(tenantId);
```
itzg, line 523: Now I really see what @jjbuchan means in https://github.com/racker/salus-telemetry-model/pull/143#issuecomment-668106710 since doing the database level deleteAll bypasses any events to trigger unbinding operations. While the net result via the overarching admin API call is pretty much the same, these individual calls could be useful for QE test cleanup.

harageth: I thought this was to easy. I have updated this with sending that event after all of them have been deleted. Its very similar to what is happening for the agent installs.

PR #87, opened 2020-08-05
Added validation support in ConfigurationProperties class

rohitsngh27:
# Resolves
https://jira.rax.io/browse/SALUS-977

# What
Added validation support in ConfigurationProperties class.


itzg:
Don't forget to add @harageth and @shabd67 as reviewers. (I already added them on this one)

src/main/java/com/rackspace/salus/resource_management/config/ResourceManagementProperties.java
```diff
@@ -1,16 +1,20 @@
...
 @ConfigurationProperties("salus.resource-management")
 @Component
 @Data
+@Validated
 public class ResourceManagementProperties {
 
   /**
    * Specifies the number of rows fetched per roundtrip when querying metadata keys.
    */
+  @Min(0)
```
itzg, line 18: Seems like zero would be invalid so let's go with
```suggestion
  @Min(1)
```

PR #97, opened 2020-12-10
fixed - Search by resource ID with space fails with 500

shabd67:
# Resolves

[SALUS-1073](https://jira.rax.io/browse/SALUS-1073)

# What

Search by resource ID with space fails with 500

# How

Added URL encoding for resourceId and then decoding at resource-management app level

# Depends on

https://github.com/racker/salus-telemetry-api/pull/114

### DIFF
--- a/src/main/java/com/rackspace/salus/resource_management/web/controller/ResourceApiController.java
+++ b/src/main/java/com/rackspace/salus/resource_management/web/controller/ResourceApiController.java
@@ -110,7 +110,6 @@ public class ResourceApiController {
   @ApiOperation(value = "Gets specific Resource for specific Tenant")
   public ResourceDTO getByResourceId(@PathVariable String tenantId,
       @PathVariable String resourceId) throws NotFoundException {
-
     return resourceManagement.getResourceDTO(tenantId, resourceId);
   }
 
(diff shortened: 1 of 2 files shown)


src/main/java/com/rackspace/salus/resource_management/web/controller/ResourceApiController.java
```diff
@@ -109,8 +111,9 @@ public SseEmitter getAllWithPresenceMonitoringAsStream() {
   @GetMapping("/tenant/{tenantId}/resources/{resourceId}")
   @ApiOperation(value = "Gets specific Resource for specific Tenant")
   public ResourceDTO getByResourceId(@PathVariable String tenantId,
-      @PathVariable String resourceId) throws NotFoundException {
-    return resourceManagement.getResourceDTO(tenantId, resourceId);
+      @PathVariable String resourceId) throws NotFoundException, UnsupportedEncodingException {
+    return resourceManagement.getResourceDTO(tenantId,
+        URLDecoder.decode(resourceId, "UTF-8"));
```
itzg, line 116: Similar [to my other comment](https://github.com/racker/salus-telemetry-api/pull/114#discussion_r540248616), Spring MVC should already be doing proper URL decoding, so something doesn't seem right about needing to explicitly handle it.

itzg:
Looks like this PR can be closed now since there's no files changed?

